PR #10, opened 2019-01-19
Lesson 1 - CodingBat sum13 exercise

Rockwell70:
Please review my solution to CodingBat exercise sum13 from the List 2 group. While it works I think there must be a more efficient way to achieve the same result. Below is the description of the exercise:

Return the sum of the numbers in the array, returning 0 for an empty array. Except the number 13 is very unlucky, so it does not count and numbers that come immediately after a 13 also do not count

.sum13([1, 2, 2, 1]) → 6
sum13([1, 1]) → 2
sum13([1, 2, 2, 1, 13]) → 6
--

### DIFF
--- a/.idea/Intro_Class.iml
+++ b/.idea/Intro_Class.iml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="PYTHON_MODULE" version="4">
+  <component name="NewModuleRootManager">
+    <content url="file://$MODULE_DIR$" />
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+  </component>
+</module>
--- a/.idea/Intro_Class.iml
+++ b/.idea/Intro_Class.iml
@@ -2,7 +2,7 @@
 <module type="PYTHON_MODULE" version="4">
   <component name="NewModuleRootManager">
     <content url="file://$MODULE_DIR$" />
-    <orderEntry type="inheritedJdk" />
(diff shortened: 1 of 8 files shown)


ldconejo:
Hi Jeff,

Your solution works well! Here's another possible solution:

```
def sum13alt(nums):
    # Initialize SUM to zero, so that there is always a value to return (if there is an early return)
    sum = 0
    # Lists is Python can be iterated, so no need to calculate the length of the a list 
    # before iterating through its elements. "for value in nums" can be read as "for every value in the list called nums, do the following:"
    for value in nums:
        # If the value is the 'unlucky' number, nothing left to do but return the current value of the sum variable
        if value == 13:
            return sum
        # Otherwise, add the value to the overall sum
        sum += value
    return sum
```

Hope this helps!

Regards,

Luis

Rockwell70:
Thank you Luis. It’s really valuable to learn how a professional developer thinks through these things.

Jeff

Sent from my iPad

On Jan 19, 2019, at 11:04 PM, Luis Conejo-Alpizar <notifications@github.com<mailto:notifications@github.com>> wrote:


Hi Jeff,

Your solution works well! Here's another possible solution:

def sum13alt(nums):
    # Initialize SUM to zero, so that there is always a value to return (if there is an early return)
    sum = 0
    # Lists is Python can be iterated, so no need to calculate the length of the a list
    # before iterating through its elements. "for value in nums" can be read as "for every value in the list called nums, do the following:"
    for value in nums:
        # If the value is the 'unlucky' number, nothing left to do but return the current value of the sum variable
        if value == 13:
            return sum
        # Otherwise, add the value to the overall sum
        sum += value
    return sum


Hope this helps!

Regards,

Luis

PR #12, opened 2019-01-20
Session 2 Exercises for review.

Douglas-Klos:
Here are my session 2 exercises, fizzbuzz.py, grid_printer.py, and series.py.  The fizzbuzz was very similar to one of the application questions for this course, and this time I got to do it slightly different.  If you could let me know if anything is missing, or should be done differently, I'd appreciate it.  Hopefully it's OK to work ahead, I'm having a good time learning python.

Also I've completed all the Python codingbat activities, they were quite helpful, and while I understand they're not graded, here's a link to a report of them.

https://codingbat.com/done?user=dougklos@gmail.com&tag=1954454802

Thanks

~Douglas Klos


students/douglas_klos/session2/fizzbuzz.py
```diff
@@ -0,0 +1,36 @@
...
+# January 17th, 2019
+# Python 210, Assignment 2
+# fizzbuzz.py
+
+# For loop is not inclusive, so we run it to 101 to include 100.
+def fizzbuzz(n=100):
+    """ displays fizzbuzz to specified number, defaults to 100 """
+    
+    # For loop is not inclusive, so we run it to 101 to include 100.
+    for num in range(1,n+1):
```
mdgregor: you'll want to include a space after your commas.  Python has a particular "style" that acceptable to the general community which can be found here: https://www.python.org/dev/peps/pep-0008/

Douglas-Klos: Thank you for bringing the pep8 styling to my attention, I believe the code now conforms to those standards.  From what I read, it seems I wouldn't want white space around the * operator like below, from grid_printer.py, though it looks a bit strange to me like this.  (I don't know how to link directly into the file like you did yet...)  Would this be correct?

`    str1 = '+ ' + '- '*4 + '+ ' + '- '*4 + '+ ' + '\n'`
`    str2 = '| ' + '  '*4 + '| ' + '  '*4 + '| ' + '\n'`

`    print((('| ' + '  '*y) *x) + '|')`
`    print((('+ ' + '- '*y) *x) + '+\n')`

Thanks again

mdgregor: You should also have a space between every operator `x + y * 2 / 3`.
```str1 = '+ ' + '- ' * 4 + '+ ' + '- ' * 4 + '+ ' + '\n'```

File: students/douglas_klos/session2/series.py [hunk 2]
```diff
@@ -0,0 +1,79 @@
...
+    :param n0=0: value of zeroth element in the series
+    :param n1=1: value of first element in the series
+    This way, if no arguments are passed, it defaults to a Fibonacci sequence.
+
+    This function generalizes the fibonacci() and the lucas(),
+    so that this function works for any first two numbers for a sum series.
+    """
+
+    if (n == 0):
+        return n0
```
mdgregor: Its hard here with mathematics notation making sense here, but you'll want to avoid non descriptive variable names in the future.  Or having a several variables where the only thing that differs is a number at the end, like `str1` and `str2`.  That doesn't tell me what that string is.

PR #14, opened 2019-01-20
lesson01 task01

syntaxlogofree:


students/x_y/session01/break_me.py
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/python3
+
+#NameError
+x*3
```
mdgregor: also make sure there is a space on either side of an operator

mdgregor: Here is the style guide for python that most devs follow like a religion.  https://www.python.org/dev/peps/pep-0008/

PR #15, opened 2019-01-21
added break_me.py to session01 directory

brickm:


ldconejo:
Code looks good, considering the intention is to trigger exceptions. One general suggestion, which is probably not overly relevant in this case, is to keep function definitions separated from where they're executed, then, at the end of your script, add a main execution block, where you call your functions:

'''
if __name__ == '__main__':
'''
    
This is to avoid any rogue code executing by accident later on when you have bigger scripts.

Regards,

Luis

PR #16, opened 2019-01-21
Codingbat sum67

Rockwell70:
I've completed all of the condingbat exercises except for this last one. I've been working on it for 2 days now and cannot get past it. 

I've excluded summing the resulting revised list because I want to see if my function is return a correctly modified list.

The code works only if there is one occurrence of the 6 - 7 series. I'm pretty sure a regex would fix this but codingbat doesn't allow import statements.

Any guidance on how to proceed would be much appreciated.

Jeff


Rockwell70:
This is my submission for the Lesson 02 Exercise: Grid Printer

ldconejo:
Hi Jeff,

That's an interesting one. I coded up a possible solution which passes all test cases. I'm pretty sure it could be optimized further:

```
def sum67(nums):
  stop_counting = False
  sum = 0
  for number in nums:
    # Converting to string, since it says "numbers starting with""
    # With a string, we can use the startswith string method.
    str_number = str(number)
    if str_number.startswith("6"):
      # This flag will prevent the function from adding up new values
      # until a number starting with 7 is detected
      stop_counting = True
      # Continue statement skips to the next number
      # so that the '6' is not added
      continue
    if stop_counting:
      # Only check for '7' is stop_counting == True
      if str_number.startswith("7"):
        # If the number is 7, set stop_counting to False
        stop_counting = False
      # Since we are in a stop_counting == True state, jump to the next number
      # no need to add the current number to the sum.
      continue
    # If no other conditions applied, the only thing left is to 
    # add up the number to the total sum.
    sum += number
  return sum
```

Let me know if you have any questions.

Regards,

Luis

PR #19, opened 2019-01-21
added a readme file

ejnielse:


ejnielse:
Created session01 Directory with break_me.py file

PR #20, opened 2019-01-23
Session 2 exercises: fizzbuzz & fibonacci

Rockwell70:
These are my submissions for fizzbuzz & fibonacci.


ldconejo:
The grid problem was a interesting one. As a matter of habit, I run Pylint to check the code for PEP8 issues. In the case of your implementation of _grid.py_, Pylint flags the fact that the "k" and "i" variables within _grid(a, b)_ are unused, but the reality is that the for statements there are used for repeating a piece of code, and it works well.

-Luis

ldconejo:
On the fizzbuzz exercise, I did a quick search on _elif_ versus _continue_. I found this on StackOverflow:

https://stackoverflow.com/questions/22540513/using-continue-instead-of-if-elif 

The accepted answer defends that using _elif_ makes your code more readable. I guess is a debate that we can continue to have. In any case, here's your code, using _continue_ and _if_:

```
def fizzbuzz():
    for i in range(1,101):
        if (i % 3 == 0 and
        i % 5 == 0):
            print('FizzBuzz')
            continue
        if i % 3 == 0:
            print('Fizz')
            continue
        if i % 5 == 0:
            print('Buzz')
            continue
        print(i)


if __name__ == '__main__':
    fizzbuzz()
```

-Luis

ldconejo:
For the Fibonacci exercise, the goal with the _sum_series_ function is to generalize both _fibonacci_ and _lucas_, to the point in which you can even modify those two functions to call _sum_series_. The way it is currently implemented is the other way around, with _sum_series_ calling _fibonacci_ and _lucas_. Please correct this and make sure your incorporate your _assert_ statements at the end of _series.py_

Once those fixes are made, you can submit a new pull request.

Regards,

Luis

Rockwell70:
I was a bit confused by altering sum_series so that fibonacci & lucas. I couldn't understand why a function would use another function to call itself. I'll work through it and resubmit.

Thanks for the article about continue. Really interesting. I'll also check out Pylint. I've heard of it but never given it a try.

Best,
Jeff

ldconejo:
Hi Jeff,

I like Pylint. It's recommendations can get annoying sometimes, but it acts like a very strict grammar teacher, constantly pointing out your errors. PEP8 is not always perfect, but I think that the order it brings contributes greatly to readability. Give it a try and let me know what you think!

Regards,

Luis

P.S.: There are other tools which you might like to explore, Flake8 is one that comes to mind right away.

Rockwell70:
Hi Luis,
Pylint is really good. I agree that it's a bit picky, but the discipline it brings is worth it.

Best,
Jeff

PR #21, opened 2019-01-24
Request assistance with Functions operations

awhitty1:
Hello,

I'm looking for feedback on my "grid_func.py" file. 
The python file is operational, however, it does contain several print statements that to an experience programmer would be unnecessary.  Welcome feedback to reduce the # of print statements.

-Alex Whitty


awhitty1:
Hello,

I'm looking for feedback on my "grid_func.py" file.
The python file is operational, however, it does contain several print statements that to an experience programmer would be unnecessary. Welcome feedback to reduce the # of print statements.

-Alex Whitty

mdgregor:
> Hello,
> 
> I'm looking for feedback on my "grid_func.py" file.
> The python file is operational, however, it does contain several print statements that to an experience programmer would be unnecessary. Welcome feedback to reduce the # of print statements.
> 
> -Alex Whitty

So you are correct, there is a lot of repetition here.  This would typically be solved with a nested `for` loop.  I don't believe these have been introduced yet so fine to omit for now.  You can see more information on them in Lesson 03 content or view some documentation on them here https://wiki.python.org/moin/ForLoop

students/alex_whitty/lesson2/grid_func.py
```diff
@@ -0,0 +1,62 @@
+
+def plus():
+    plus = '+'
+    return plus
```
mdgregor: This function doesn't really gain you anything as its just outputing `+`.  So on line 18 you are actually typing more by doing `plus()` than `+`

ldconejo:
Closing this one for now, as it seems the questions Alex had have been answered by Mike.

-Luis

PR #23, opened 2019-01-24
Session 3 exercises for review

Douglas-Klos:
Here are my session 3 exercises.  I'd appreciate any feedback on these, potential errors, areas that can be improved, or suggestions.  I'm having some (hopefully simple) surgery tomorrow (Friday) and likely won't be doing as much while recovering.

Thanks.

### DIFF
--- a/students/douglas_klos/session3/list_lab.py
+++ b/students/douglas_klos/session3/list_lab.py
@@ -50,7 +50,6 @@ def series2():
     the list.  Asks the user for a fruit to delete, finds it, and
     deletes it.  Bonus section: Multiplies the list by 2, continues
     asking for a fruit until one is found, then deletes all occurances.
-
     """
 
     print("Series 2:")
--- a/students/douglas_klos/session3/list_lab.py
+++ b/students/douglas_klos/session3/list_lab.py
@@ -1,0 +1,137 @@
+#!/usr/bin/env python3
+
+# Douglas Klos
+# Python 210, Exercise 3
+# January 24st, 2019
+# list_lab.py
+
+
+def series1():
+    """
+    Creates the specified list and displays it. Asks user for another
+    fruit and adds it to end of list.  Displays list again.  Asks user
+    for a number, prints it back and its coressponding fruit.
+    Adds another fruit to the beginning of list using +.
+    Adds another fruit to the beginning of list using insert.
+    Displays all fruit in list beginning with P.
+    """
+
+    print("Series 1:")
+    fruit = ["Apples", "Pears", "Oranges", "Peaches"]
+    print(fruit)
+
+    # Ask a user for a fruit and add to end of list.
+    fruit += (input("Enter the name of fruit to be added: "),)
+    print(fruit)
+
+    # Ask user for anumber and display the number back to the use and
+    # the fruit corresponding to that number.  No error trapping.
+    index = input("Enter the index of fruit to print: ")
+    print("index = " + index + ", fruit = " + fruit[int(index)-1])
+
+    # Add fruit to beginning using the "+".
+    fruit = [input("Enter fruit to add at beginning of list: ")] + fruit
+    print(fruit)
+
+    # Add fruit to beginning using insert().
+    fruit.insert(0, input("Enter fruit to insert at beginning of list: "))
+    print(fruit)
+
+    # Display all the fruits that begin with "P", using a for loop.
+    for i in range(len(fruit)):
+        if fruit[i][0] == "P":
+            print(fruit[i] + " ", end="")
+    print("\n")
+
+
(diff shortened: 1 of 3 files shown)


students/douglas_klos/session3/list_lab.py
```diff
@@ -0,0 +1,136 @@
...
+    print(fruit)
+
+    # Ask a user for a fruit and add to end of list.
+    fruit += (input("Enter the name of fruit to be added: "),)
+    print(fruit)
+
+    # Ask user for anumber and display the number back to the use and
+    # the fruit corresponding to that number.  No error trapping.
+    index = input("Enter the index of fruit to print: ")
+    print("index = " + index + ", fruit = " + fruit[int(index)-1])
```
mdgregor: consider using an `f` string instead of using `+` for creating strings
`print(f"index = {index}, fruit = {fruit[int(index) - 1]}")`  Makes it a little more readable

students/douglas_klos/session3/list_lab.py
```diff
@@ -0,0 +1,136 @@
...
+
+    # Add fruit to beginning using the "+".
+    fruit = [input("Enter fruit to add at beginning of list: ")] + fruit
```
mdgregor: Try to use the built in functions of your list object such as `.insert` to insert a value at a given index
`fruit.insert(0, input("Enter fruit to add at beginning of list: "))`
I would probably also break out the `input` command to put into a variable instead of inline with the `.insert` statement just to make it more readable.

students/douglas_klos/session3/list_lab.py
```diff
@@ -0,0 +1,136 @@
...
+    print(fruit)
+
+    # Add fruit to beginning using insert().
+    fruit.insert(0, input("Enter fruit to insert at beginning of list: "))
```
mdgregor: lol, I guess I should have read a little more before commenting above :)

students/douglas_klos/session3/list_lab.py
```diff
@@ -0,0 +1,136 @@
...
+    print(fruit)
+
+    # Display all the fruits that begin with "P", using a for loop.
+    for i in range(len(fruit)):
+        if fruit[i][0] == "P":
+            print(fruit[i] + " ", end="")
```
mdgregor: Again take advantage of the built in functions of the list object.  You also don't need to get a `len` of the list to iterate through it.
```
for i in fruit:
    if i.startswith("P"):
        print(f"{fruit[i]} ", end="")
```

students/douglas_klos/session3/list_lab.py
```diff
@@ -0,0 +1,136 @@
...
+    while(not fruit_found):
+        print(fruit)
+        kill_fruit = input("Enter fruit to remove from list: ")
+        for i in range(len(fruit)):
+            if fruit[i] == kill_fruit:
+                fruit_found = True
+                remove_tuple += (fruit[i],)
+        for item in remove_tuple:
+            fruit.remove(item)
+    print(fruit)
```
mdgregor: This could be simplified quite a bit.
```
kill_fruit = input("Enter fruit to remove from list: ")
while kill_fruit in fruit:
    fruit.remove("kill_fruit")
```

students/douglas_klos/session3/list_lab.py
```diff
@@ -0,0 +1,136 @@
...
+    for item in fruit:
+        answer = input("Do you like " + item.lower() + "? ")
+        while answer != 'no' and answer != 'yes':
+            print("Please enter yes or no")
+            answer = input("Do you like " + item.lower() + " :")
+        if answer == "no":
+            remove_tuple += (item,)
+    for item in remove_tuple:
+        fruit.remove(item)
+    print(fruit)
```
mdgregor: Lets also simplify this a bit
```
for item in fruit[::]:  # We iterate through a copy of the list because we are making edits to the original
   answer = input(f"Do you like {item}?: ")
    if answer == "no":
        fruit.remove(item)
```

students/douglas_klos/session3/list_lab.py
```diff
@@ -0,0 +1,136 @@
...
+    deleted.  For any answer this is not yesor no, the user is
+    to re-enter an answer, yes or no.
+    """
+
+    print("Series 3:")
+    fruit = ["Apples", "Pears", "Oranges", "Peaches"]
+    remove_tuple = ()
+    for item in fruit:
+        answer = input("Do you like " + item.lower() + "? ")
+        while answer != 'no' and answer != 'yes':
```
mdgregor: a more "pythonic" way to check this is `if answer.lower() not in ('no', 'yes'):`

PR #25, opened 2019-01-25
ScottL - Assignment02 File Submission

sdlipp:
Submitting 3 files for review to complete the requirements for Assignment02.  The following files are located in the session02 folder:

- series.py

- strings.py

- fizzbuzz.py


mdgregor:
Can you remove the `.idea/` from your PR.  You'll want to add `.idea/` to your `.gitignore` file so it doesn't try to include it in the commits.  Good job on everything else, only comment I have is that you'll want to use `snake_case` for all functions and variables.  Only class names use `CamelCase`

mdgregor:
Hey Scott.  Going to close this for now until you can get the `.idea/` folder sorted out.

sdlipp:
I tried removing the .idea folder before I made the push and followed the prompts from git as best as I could.  I also have the following in the gitignore file:

# IDE
*.idea   --this was existing
.idea/   --i manually added this to match your comment from above

Is there something else that I still need to do?

PR #26, opened 2019-01-26
Submitting Lesson 02 Exercises

jroam142:
All three lesson 02 exercises are in the lesson02_exercises folder.


ldconejo:
Good work on all three assignment, Jerome. On _series.py_, since you have _return_ statements, you can get rid of _else_ and _elif_. The function is currently written like this:

```
def sum_series(n, o=0, p=1):
    """ Returns the integer at index n from a series based on the starting
        numbers specified with parameters o and p. The default settings of these
        parameters result in the fibonacci series """
    if n <= 0:
        return o
    elif n < p + 1:
        return n
    else:
        return sum_series(n - 1, o, p) + sum_series(n - 2, o, p)
```

It would work the same by replacing _elif_ with _if_ and removing _else_:

```
def sum_series(n, o=0, p=1):
    """ Returns the integer at index n from a series based on the starting
        numbers specified with parameters o and p. The default settings of these
        parameters result in the fibonacci series """
    if n <= 0:
        return o
    if n < p + 1:
        return n
    return sum_series(n - 1, o, p) + sum_series(n - 2, o, p)
```

Regards,

Luis

PR #27, opened 2019-01-26
Pull request

UncleanlyCleric:
Submitting even though I wiped a lot of commit messages accidentally.  I know this will reflect in my scare, but I accept that.


students/jesse_miller/session02/fizzbuzz.py
```diff
@@ -0,0 +1,10 @@
+#!/usr/local/bin/python3
+for i in range(1,101):
+    string = ""
+    if i % 3 == 0:
+        string = string + "Fizz"
```
mdgregor: you can use `string += "Fizz"`

students/jesse_miller/session02/fibonacci_test.py
```diff
@@ -0,0 +1,19 @@
+#!/usr/local/bin/python3
+"""This is a test to make sure I understand how the code works."""
+import sys
+"""import sys for .argv input"""
+n = int(sys.argv[1])
```
mdgregor: You shouldn't use `sys.argv` for these assignments as taking in user input and validating it and so forth is a good learning experience.

students/jesse_miller/session02/print_grid.py
```diff
@@ -0,0 +1,31 @@
...
+
+def print_row():
+    """This is where the magic starts.  It prints one line and four columns"""
+    print_line()
+    do_four(print_post)
+
+def print_grid():
+    """The remainder of the magic.  It doubles the above function and prints the last row"""
+    do_two(print_row)
+    print_line()
```
mdgregor: This is a bit to complex to follow and difficult to read and understand what is going on.  Take a look at `for` loops which I think would help you achieve what you are trying to do here.
```
print_line()
for line in range(3):
    for post in range(3):
        print_post()
    print_line()
```
see if you can update this file to reflect using for loops

UncleanlyCleric: I will try, I'm not certain I can get it done by tonight but I will do what I can.

PR #32, opened 2019-01-27
Submitting Lesson (2) Assignments

awhitty1:
Submitting lesson (2) assignments:
fizzbizz.py
grid_func.py
grid_printer.py
print_grid2.py
series.py


students/alex_whitty/lesson2/print_grid2.py
```diff
@@ -0,0 +1,51 @@
...
+    print(plus() + minus(1, 1) + plus() + minus(1, 1) + plus())
+    print(vert_dash(2, 2))
+    print(plus() + minus(1, 1) + plus() + minus(1, 1) + plus())
+    print(vert_dash(2, 2))
+    print(plus() + minus(1, 1) + plus() + minus(1, 1) + plus())
+
+
+
+
+print_grid(2,1)
```
mdgregor: the variables your passing aren't actually being used in your program.  You should swap out your `2`s and `1`s for the variables.

students/alex_whitty/lesson2/print_grid2.py
```diff
@@ -0,0 +1,51 @@
...
+Tried the nested for loop to reduce the number of print statements.
+However, I could not get the align correct to complete all four sides
+of the grid at once during execution. 
+
+
+
+
+def print_grid(v, m):
+    for i in range(2):
+        for i in range(1):
```
mdgregor: You can't use the same `i` var in nested loops as its still in scope, so you are overriding the first loop's `i` variable.

mdgregor: nvm, looks like you can, thought i wouldn't recommend it :)

PR #36, opened 2019-01-28
Added Session02 Files

ejnielse:


ldconejo:
Merging pull request. All three exercises have already been graded by TA.

PR #39, opened 2019-01-28
adding three python files for lesson 2

brickm:


ldconejo:
Hi Brick,

For the grid assignment, the goal is that you use multiple values passed to the _print_ statement, as well as string multiplication and a function with two parameters to generate the grid. The way it is currently implemented takes fixed values to generate the grid. Whenever you have some time, take another shot at this assignment and submit a new pull request with a new implementation.

Thanks,

Luis

PR #41, opened 2019-01-28
Pull request to for module 2 homework files

bbirsoz1:


ldconejo:
Hi Begun,

The code for the fizzbuzz problem looks good. This is the first case I've seen in which someone used mod 15 to determine a number is divisible by both 3 and 5, very clever!

From a formal perspective, some things you can change in how you write your code:

if i % 15==0: 

Always try to a space before and after a comparison operator, the line above should look like this:

if i % 15 == 0:

For this one:

print (i)

Avoid leaving a whitespace between the _print_ statement and the parenthesis.

You can also add a docstring at the beginning of your script to explain what it does. For example:

'''
This is the solution to the fizzbuzz exercise
'''
Keep up the good work!

Luis

ldconejo:
Your Fibonacci / Lucas implementation is also very good. Some things to consider:

```
def fibonacci(n):
    if n < 0:
        return None
    elif n == 0:
        return 0
    elif n == 1:
        return 1
    else:
        return fibonacci(n-2) + fibonacci(n-1)
```

Note that every condition returns (i.e., exits execution of the function _fibonacci_) if True, so you don't really need _elif_, they can all be replaced with _if_. The same goes for the final _else_:

```
def fibonacci(n):
    if n < 0:
        return None
    if n == 0:
        return 0
    if n == 1:
        return 1
    return fibonacci(n-2) + fibonacci(n-1)
```

PR #43, opened 2019-01-28
Assignment 2

woosukjeung:


ldconejo:
Hi Wookseok,

Your code looks good. For the _print_ statement, avoid leaving a space between it and its parentheses:

print ("FizzBuzz") --> print("FizzBuzz")

-Luis

ldconejo:
In the Fibonacci exercise, you don't need an _elif_ statement after a return, since the code after the return will only run if the condition is False. Additional to that, you don't need a _pass_ statement at the end of the function (usually, the _pass_ statement is only a placeholder for pending code:

```
def lucas(n):
    if n == 0:
        return 2
    elif n == 1:
        return 1
    return lucas(n-2) + lucas(n-1)
    pass
```

Should be:

```
def lucas(n):
    if n == 0:
        return 2
    if n == 1:
        return 1
    return lucas(n-2) + lucas(n-1)
```

PR #44, opened 2019-01-29
session3 slicing_lab

Rockwell70:
This is my pull request for session3 slicing_lab.


Rockwell70:
These are my list_lab and strformat_lab submissions.

students/jeff_shabani/session03/list_lab.py
```diff
@@ -0,0 +1,65 @@
...
+    result = []
+    for i in iterable:
+        result.append(i[::-1])
+    iterable.pop()
+    print(f'Copy of original with letters of each item reversed: {result}')
+    print(f'Original with last item removed: {iterable}')
+
+
+if __name__ == '__main__':
+    series_one(FRUIT_LIST)
```
mdgregor: I would suggest initializing each fruit list in each function rather than using this comment out approach.

students/jeff_shabani/session03/list_lab.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python3
+
+
+FRUIT_LIST = ['Apples', 'Pears', 'Oranges', 'Peaches']
+
+
+def series_one(iterable):
+    print(iterable)
+    answer = str(input('What fruit to you want to add?'))
```
mdgregor: no need to cast to a `str`. it will already be one.

students/jeff_shabani/session03/list_lab.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python3
+
+
+FRUIT_LIST = ['Apples', 'Pears', 'Oranges', 'Peaches']
+
+
+def series_one(iterable):
+    print(iterable)
```
mdgregor: I know that your just passing in the fruit list, but `iterable` doesn't describe what the object is at all.  Just call it `fruit_list` here.

students/jeff_shabani/session03/list_lab.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def series_three(iterable):
+    acceptable_answers = {'yes', 'no'}
+    result = []
+    for i in iterable:
+        answer = input(f'Do you like {i}?')
+        while answer.lower() not in acceptable_answers:
+            answer = input(f'Please answer yes or no: Do you like {i}?')
```
mdgregor: need a `continue` after this

PR #50, opened 2019-01-30
Session03 - Lessons

pauleclifton:


students/paul_c/session03/list_lab.py
```diff
@@ -0,0 +1,89 @@
...
+# Series 1
+print("\n*** Series 1 ***\n")
+fruits = ["Apples", "Pears", "Oranges", "Peaches"]
+print(fruits)
+user_fruit = input("Please type in a new fruit > ")
+fruits.append(user_fruit)
+print(fruits)
+user_index = input("Please enter a number from 0 - 4 > ")
+user_selected = fruits[int(user_index)]
+print("You have selected fruit number", user_index, '"'+user_selected+'"', "from the list.")
```
mdgregor: consider and `f` string for this: `print(f"You have selected fruit number {user_index} {user_selected} from the list.")`  Much easier to read.

students/paul_c/session03/list_lab.py
```diff
@@ -0,0 +1,89 @@
...
+pineapples = ["Pineapples"]
+fruits = pineapples + fruits
+fruits.insert(0, "Papayas")
+for i in fruits:
+    if i.startswith("P|p"):
```
mdgregor: `if i.lower().startswith("p"):`

students/paul_c/session03/list_lab.py
```diff
@@ -0,0 +1,89 @@
...
+    if user_peaches != 'yes' and user_peaches != 'no':
+        print("Please enter yes or no.")
+        continue
+    elif user_peaches == "no":
+        fruits.remove('Peaches')
+        break
+    else:
+        break
+
+print(fruits)
```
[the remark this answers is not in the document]
pauleclifton: Wow, a for loop would have made things much easier. I wrote so much duplicate code. Thanks for the feedback. In the future, I will considering iterating though sequences before duplicating all that code.

students/paul_c/session03/list_lab.py
pauleclifton: I need to get a better handle on the use of fstrings. They seem super useful.

students/paul_c/session03/list_lab.py
pauleclifton: Cool, I didn't know you could do it that way. Thanks!

PR #53, opened 2019-01-31
submitting session03 homework

yiyunxu:


students/elaine_x/session03/list_lab_ex.py
```diff
@@ -0,0 +1,112 @@
...
+#Python 210
+#Session 03 - List Lab
+#Elaine Xu
+#Jan 28,2019
+###########################
+'''
+
+############# Series 1 ################################
+print("Series 1\nList of Fruit:")
+FRUITS = ['Apples', 'Pears', 'Oranges', 'Peaches']
```
mdgregor: since this list is changing, you don't want to capitalize it, as that signifies that its a constant (i.e. immutable or unchanging)

students/elaine_x/session03/list_lab_ex.py
```diff
@@ -0,0 +1,112 @@
...
+print(FRUITS)
+
+#Ask the user for another fruit and add it to the end of the list
+FRUIT = input('Enter another fruit: ')
+FRUITS.append(FRUIT)
+print(FRUITS)
+
+#Ask user for a number, display the number and the fruit corresponding to the number
+NUMBER = int(input('Enter an index number (1 is first): '))
+print("The "+ str(NUMBER) +"th fruit is "+FRUITS[NUMBER-1])
```
mdgregor: don't need to cast to `str` to print here

students/elaine_x/session03/list_lab_ex.py
```diff
@@ -0,0 +1,112 @@
...
+
+#Add another fruit to the beginning of the list using “+” and display the list
+print()
+print('Adding "Kiwi" to the beginning of the list')
+FRUITS = ['Kiwi'] + FRUITS
```
mdgregor: better to use `fruits.insert(0, 'Kiwi')`

students/elaine_x/session03/list_lab_ex.py
```diff
@@ -0,0 +1,112 @@
...
+print()
+print('Adding "Banana" to the beginning of the list')
+FRUITS.insert(0, 'Banana')
+print(FRUITS)
+
+#Display all the fruits that begin with “P”, using a for loop
+print()
+print('Display all the fruits that begin with "P"')
+for item in FRUITS:
+    if item[0] == "P":
```
mdgregor: `if item.lower().startswith("p"):`  If they want it to be case sensitive, drop the `.lower()`

students/elaine_x/session03/list_lab_ex.py
```diff
@@ -0,0 +1,112 @@
...
+print()
+print("Series 3\nList of Fruit:")
+FRUITS3 = FRUITS[:]
+print(FRUITS3)
+FRUITS3_COPY = FRUITS3[:]
+
+for fruit in FRUITS3:
+    choice = ''
+    while choice.lower() != 'yes' and choice.lower() != 'no':
+        choice = input("Do you like " + fruit.lower() + "? (yes or no):")
```
mdgregor: if they just enter `blah` it will move on to the next fruit, how can you avoid that?

yiyunxu: while choice.lower() != 'yes' and choice.lower() != 'no':
The while loop will keep looping until the input is 'yes' or 'no' (upper and lower case).

mdgregor: lol, yep your right, read it wrong.

students/elaine_x/session03/list_lab_ex.py
yiyunxu: Thank you mike for the comments. they are very helpful. When you are done reviewing, should I close this request and re-submit.

students/elaine_x/session03/mailroom.py
```diff
@@ -0,0 +1,104 @@
...
+            donor_db[fullname_index][1][0] = donor_db[fullname_index][1][0] + donation_amount
+            donor_db[fullname_index][1][1] = donor_db[fullname_index][1][1] + 1
+            print(f"{donation_amount} has been added to {fullname}'s donation history.")
+            #print(donor_db) #for testing
+            break
+    #send thankyou note
+    print("Composing Thank You email:")
+    print(f'Thank you {fullname} for your generous donation of {donation_amount:^10.2f}!')
+    print()
+    menu()
```
mdgregor: You don't want to call the `menu()` function here as you'll get into a weird recursion thing when trying to exit.  You can just leave this line out and it will return to your `while True` loop for your menu.

students/elaine_x/session03/mailroom.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def sort_total_donation(number):
+    '''sort donors by total historical donation amount'''
+    return number[1][0]
+
+
+def main():
+    '''main operation of the program'''
+    menu()
```
mdgregor: move this to inside the `while True` loop and you won't have to call it a bunch of times in the other functions.

students/elaine_x/session03/mailroom.py
mdgregor: nvm, don't think this will do recursion since its only a text print out

students/elaine_x/session03/list_lab_ex.py
mdgregor: nope, any changes you make automatically get pulled into the request.  But i've merged this PR anyway so you can edit for the next one if you like.

PR #57, opened 2019-02-02
mailroom submission

Rockwell70:
This is my mailroom submission for session 3


students/jeff_shabani/session03/list_lab.py
```diff
@@ -3,63 +3,69 @@
 
 FRUIT_LIST = ['Apples', 'Pears', 'Oranges', 'Peaches']
```
mdgregor: also remember that this isn't a constant so it shouldn't be in all caps.  Its a constantly changing list.  Also constants should use an immutable data structure like a tuple rather than a list.

students/jeff_shabani/session03/mailroom.py
```diff
@@ -0,0 +1,71 @@
...
+        if response == '1':
+            get_a_name()
+        elif response == '2':
+            create_report()
+        elif response == '3':
+            quit_the_program()
+        else:
+            print('Please make a valid selection')
+
+main()
```
mdgregor: i would suggest using the `if __name__ == '__main__':` here as that is the proper convention rather than just calling `main()`

students/jeff_shabani/session03/mailroom.py
```diff
@@ -0,0 +1,71 @@
...
+    for k,v in DONORS.items():
+        print(k)
+
+def add_new_donor(name, donor_list):
+    amount = int(input(f'How much would this donor like to donate?'))
+    donor_list[name] = [amount]
+
+
+
+def get_a_name():
```
mdgregor: I would rename this to something that describes what it does.  `get_a_name` isn't it :)

students/jeff_shabani/session03/mailroom.py
```diff
@@ -0,0 +1,71 @@
...
+        answer = input('Please enter a donor Full Name.')
+    amount = int(input('How much would this donor like to donate?'))
+    if answer not in DONORS:
+        DONORS[answer] = [amount]
+    else:
+        for k,v in DONORS.items():
+            if k.lower() == answer.lower():
+                v.append(amount)
+    print(f'Thank you {answer.split()[0]} for you generous donation of ${amount:,.0f}')
+    print()
```
mdgregor: So I re wrote it a bit, also notice the spacing between the lines makes it more readable instead of text on every line
```
def add_donations_and_send_thank_you():

    while True:

        answer = input('Please enter a donor Full Name.')

        if answer.lower() == 'list':
            view_donor_names()
            continue

        amount = int(input('How much would this donor like to donate?'))

        if answer not in DONORS:
            DONORS[answer] = [ammount]

        else:
            for name, donations in DONORS.items():
                if name == answer:
                    donations.append(amount)

        print(f'\nThank you {answer.split()[0]} for you generous donation of ${amount:,.0f}\n')
        break
```

PR #59, opened 2019-02-02
Session3 mailroom.py submission

Douglas-Klos:
The more I stare at this the more I tweak and add to it, think I just need to be finished with it.  Unless I still missed something. hmm.....

Thanks


students/douglas_klos/session3/mailroom.py
```diff
@@ -0,0 +1,203 @@
...
+        # Check to see if the name entered exists in the database
+        # If name not found, verify you want to add it.
+        elif not any(name_input in sub_list for sub_list in mailroom_db):
+            add_check = input('Donor not found, would you like to add?: ')
+            if add_check.lower() in ('y', 'yes'):
+                add_donor(name_input)
+
+        # The name was found in the database
+        else:
+            send_thank_you(name_input)
```
mdgregor: I think this function and the ones it calls make this a little to complex.  Perhaps try something like this:
```
def thank_you_menu():
    name_input = input(THANK_YOU_PROMPT):

    if name_input.lower() in ('q', 'quit'):
         return
    elif name_input.lower() in('l', 'list'):
         display_database()

    for donor in mailroom_db:
        if name_input == donor[0]:
            send_thank_you(name_input)
            return

    donation_amount = float(input("Enter donation amount: "))

    mailroom_db.append((name_input, [donation_amount]))
```
I believe the above should work and you can remove a few functions, and then within `send_thank_you` you can append a donation to the end of a donor using the same `for` loop logic i have above.

students/douglas_klos/session3/mailroom.py
```diff
@@ -0,0 +1,203 @@
...
+    """ Mailroom main function """
+
+    while True:
+        selection = input(MAIN_PROMPT)
+        if selection in '1':
+            thank_you_menu()
+        elif selection in '2':
+            create_report()
+        elif selection in '3':
+            add_donor('')
```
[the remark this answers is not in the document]
Douglas-Klos: Would something like this be more appropriate?

`    selection = input(MAIN_PROMPT)

    while selection.lower() not in ('q', 'quit'):

        if selection == '1':
            thank_you_menu()
        elif selection == '2':
            create_report()
        elif selection == '3':
            add_donor('')
        elif selection == '4':
            add_donation()
        elif selection.lower() in ('5', 'p', 'print'):
            display_database()
        else:
            print("Not a valid entry")

        selection = input(MAIN_PROMPT)`

I'm not sure how else to do a menu without a switch statement, or perhaps a dictionary with function names as values, which might be next weeks.

students/douglas_klos/session3/mailroom.py
Douglas-Klos: I did clean up my code a bit in that area, getting rid of some of the function calls, and cleaning out 20+ lines of code in the process.  I wanted to maintain the error trapping though to check for invalid inputs from the user.

Douglas-Klos:
Should I submit another pull request with the changes I've made, or just submit my new version via canvas?

Thank you for your time and assistance.

ldconejo:
Hi Doug,

Good job on this one. For future submissions, consider removing unnecessary else/elif statements placed after a _return_ statement:

```
                    if donation_input in donations:
                        print(THANK_YOU_NOTE.format(donor, donation_input))
                        return
                    else:
                        print(f'Donation from {donor} '
```

Should be:

```
                    if donation_input in donations:
                        print(THANK_YOU_NOTE.format(donor, donation_input))
                        return
                    print(f'Donation from {donor} '
```

Regards,

Luis

PR #61, opened 2019-02-03
Session03 exercises

mjchang14:
Ran into loop issues on series 3 in list_lab and the formatting for report for mailroom. I still feel like I don't fully understand nested lists (continuing to practice though), and some of my indexing code essentially came from guessing and trying a lot.


ldconejo:
Hi Marisa,

For _list_lab.py_ everything works great. One improvement you might consider in the future is to simplify the condition used by your while loop, to include all valid options in a single tuple. You are using this:

```
    while tasty != "no" and tasty != "yes":
        tasty = input("Please enter yes or no: ")
```

A simpler way to do it would be this:

```
    while tasty not in ("no", "yes"):
        tasty = input("Please enter yes or no: ")
```

Note that in this fashion, you could even define a tuple with more options with minimum coding effort.

-Luis

ldconejo:
Hi Marissa, 

Good job with the String Formatting Exercise. For future submissions, please remember to include some sample execution of the functions you create, as that speeds up the process of reviewing your code. 

To improve readability, avoid lines longer than 100 characters in your code.

Regards,

Luis

ldconejo:
Hi Marisa,

Good job on the Slicing Lab.

-Luis

PR #63, opened 2019-02-03
Add mailroom program to session folder

pauleclifton:
I am still having issues with the report section. I can't seem to get the columns to line up just right.

### DIFF
--- a/students/paul_c/session03/mailroom.py
+++ b/students/paul_c/session03/mailroom.py
@@ -1,0 +1,109 @@
+#!/usr/local/bin/python3
+
+import sys
+
+donors = {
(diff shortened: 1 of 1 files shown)


ldconejo:
Hi Paul,

Your mailroom code works correctly. For the issue with lining up your columns (I'm assuming you want to line the columns with money values to the right, do this on line 26:

        donor_row = " ".join(["{:20s} {:>20s} {:>20s} {:>20s}"]).format(*row_format)

Regards,

Luis

pauleclifton:
Thanks for the suggestion, Luis. It looks much better now.

PR #66, opened 2019-02-04
Submitting Lesson 3 files

awhitty1:
Submitting Lesson 3 files.


awhitty1:
Week self-review:
I need to work more on my if/else condition statements within the script/program workflow. At this point, I'm not sure when to use Booleans or Iteration statements while coding(if that makes sense).

ldconejo:
Hi Alex,

For the List Lab, in series 1, your code for searching fruits starting with "p" needs a couple of changes (you were on the right track, though);

for i in fruits:
    if fruits[0:] == "P": 
      print("Found a match!")
    else:
      print(i)

_fruits[0:]_ the colon will cause this to return all elements of the list, starting with the selected index. Since you're using index 0, it will simply return the entire list. I assume you want to check the first character of each fruit and see if it matches "P", so you would need something like this:

for i in fruits:
    if i[0] == "P": 
      print("Found a match!")
    else:
      print(i)

_for i in fruits_ could be read as "for each _i_ in fruits, do the following". It would also be more readable if instead of using "i" you used something more descriptive, such as:

for fruit in fruits:
    if fruit[0] == "P": 
      print("Found a match!")
    else:
      print(fruit)

Another issue is that you're trying to match a capital "P", but the fruits in your list are all lowercase. You could either correct it to look for lowercase "p" or force _fruit_ to uppercase:

for fruit in fruits:
    if fruit[0].upper() == "P": 
      print("Found a match!")
    else:
      print(fruit)

Regards,

Luis

ldconejo:
Nice touch on adding a menu to series 3!

ldconejo:
For series 4, also of the list lab, you need to reverse the order of the characters in the list. For example, "apples" would become "selppa". Please correct these issues whenever you have some time.

awhitty1:
Hello Luis,
Thank you for your feedback. I have made the recommended fixes and re-submitted via Canvas. Also, committed updates to Github.

ldconejo:
Hi Alex,

For the String Formatting Exercise, task 3 requires adding the ability to deal with a tuple of arbitrary size (for example, 7 elements). For task 4, you need to be able to detect single-digit numbers and print them with a preceding "0" (for example, "4" turns into "04"). Please make sure you add those two changes. Everything else looks good.

-Luis

ldconejo:
Hi Alex,

For the slicing lab, please reorganize your code so that all code is contained within your functions, the way it is written right now makes it very difficult to follow. For example:

```
my_list = [0, 1, 2, 3, 4, 5, 6, 7]
new_list = my_list[-1:8] + my_list[1:7] + my_list[:1]

def exchange_first_last(seq):
    return new_list
```

This case, the only thing your function does is return _new_list_, which is already defined outside of the function. _exchange_first_last()_ is also taking a parameter, _seq_ which is not being used at all. A better way to write it would be:

```
def exchange_first_last(input_list):
    new_list = input_list[-1:8] + my_list[1:7] + my_list[:1]
    return new_list
```

Then, you can call it as _exchange_first_last(my_list)_

The same applies for all the other functions in your code. You should not have lose lines of code mixed with function definitions, or functions requiring parameters that they end up not using.

-Luis

awhitty1:
Hello Luis,

Thank you for your feedback on the Slicing and String Formatting labs. I have made the recommended changes. I was using the Sublime text editor. However, I find that the PyCharm IDE displays my coding errors more vividly compared to Sublime.  Also, I see your point with the function coding workflow. Unfortunately for me, this course is the only coding exposure I have access too at present. That being said, I do welcome all your feedback.

ldconejo:
Hi Alex,

I understand. This course is, I like to think, a good place to be exposed to coding for the first time, so I hope my feedback will add value to you (I also hope you tell me if it at some point it doesn't :-)). Regarding IDEs, they do provide good help in highlighting coding issues as you write your code.

Regards,

Luis

ldconejo:
For the mailroom assignment, several changes are needed:

1. Convert your _donor_db_ to a dictionary. Something like this:

```
donor_db = {"William Gates, III": [653772.32, 12.17],
             "Jeff Bezos": [877.33],
            "Paul Allen": [663.23, 43.87, 1.32],
            "Mark Zuckerberg": [1663.23, 4300.87, 10432.0],
            "Elon Musk": [2263.23, 3300.87, 15432.0],
}
```

2. Changes to _thank_you()_:

- When the option selected is _list_, iterate over _donor_db_ as a dictionary, printing only the names of the donors, instead of donors and donations.
- Also, separate the flows for _add_donor()_ and _current_donor()_

```
def thank_you():

    answer = input("Please enter Full Name.>>>")
    if answer.lower() == 'list':
        #print(sort_key(donor_db))
        for donor in donor_db.keys():
            print(donor)
    #elif answer.lower() != 'list':
    else:
        if answer in donor_db:
            current_donor()
        else:
            add_donor()
```
3. Changes to _current_donor()_

- Change it to take an parameter with the donor name (_donor_).
- Append (similar to what you're already doing) the new amount to the list associated with the donor:

```
def current_donor(donor):
    amount = input("Please enter donor amount.>>>")
    donor_db[donor].append(amount)
```
Let me know if you have any questions about these changes. You can submit a new pull request once the changes are ready and we'll update your grade for the assignment.

Regards,

Luis

ldconejo:
Hi Alex,

For your updates to the String Formatting Lab, task 4 looks good. Task 3 would still need to be updated to take a tuple with an arbitrary size.

I will close this pull request for now, please submit a new one with any updates to your pending items on lesson 3 and potentially with some of your work for lesson 4.

Regards,

Luis

awhitty1:
Hello Luis,
I made the changes to the mailroom.py for review. However, each function in my program now requires 'one parameter argument' which I don't think is right. Also,  my append statements under def add_donor(): is producing an "AttributeError". The append method I thought was used to add data to the end of the existing dictionary without specifying the Key:Value.

awhitty1:
Hello Luis,
I figured out the append statement. Replaced with donor_db[new_donor] = [donor_amount].

awhitty1:
Hello Luis,
I just re-submitted mailroom.py with updated changes for review/grade update.

PR #69, opened 2019-02-04
Adding for Lesson 3

woosukjeung:
having a bit of problem with the List Exercise where the computer would freeze upon loading causing the delay in submission cause i had to fix my computer first

### DIFF
--- a/students/WooseokJ/List+WJ.py
+++ b/students/WooseokJ/List+WJ.py
@@ -1,0 +1,56 @@
+
+# coding: utf-8
+
+# In[10]:
+
+
+# Series 1
+print("Series 1")
+fruits = ["Apples", "Pears", "Oranges", "Peaches"]
+print(fruits)
+newfruit = input("Please type in a new fruit ")
+fruits.append(newfruit)
+print(fruits)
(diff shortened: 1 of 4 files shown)


woosukjeung:
added fixed list assignment

ldconejo:
Hi Wooseok, 

For the List Lab, when you're locating items in the fruit list by number, one of the requirements is that you consider the first element of the list of fruits to be index 1, while Python indexes lists by starting at 0. Please correct that.

For Series 2, also in the List Lab, you're asking the user to select a fruit to delete but you're not executing the deletion.

Regards,

Luis

ldconejo:
Hi Wooseok,

Great work on the string formatting exercise! One thing to consider is to use enumeration when going through the elements of a list or tuple. For example:

```
    for i in range(len(tuple_to_format)):
        print('%10s, %3d,  $%9.2f' % tuple_to_format[i])
```

Could be turned into this:

```
    for element in tuple_to_format:
        print('%10s, %3d,  $%9.2f' % element)
```

Regards,

Luis

ldconejo:
All assignment requirements met for the Slicing Lab.

PR #70, opened 2019-02-04
Session 4,5 mailroom2, trigrams, except_exercise for review / feedback

Douglas-Klos:
trigrams.py working using a recursive solution.
mailroom.py supports file writing and databases where I felt they were applicable.
session 5 except_exercise.py seems to be working also.

I appreciate any feedback / suggestions you could give me on these assignments.


students/douglas_klos/session3/mailroom.py
```diff
@@ -102,17 +72,39 @@ def add_donor(name_input):
...
+        if name.lower() in ('q', 'quit'):
+            return
+        elif not any(name in sub_list for sub_list in mailroom_db):
+            print('Donor not found, please select from list')
+        else:
+            break
+
+    while True:
+        try:
+            donation_input = input('Please enter a donation: ')
```
mdgregor: cast as a float here and you don't have to do it every time you use it.

Douglas-Klos: Thank you for the suggestion, however I was looking for feedback on session4 mailroom2 (which has the dictionary changes) and trigrams. That said I was able to make the float cast change in the session4 file as well, but there's a fair amount different to review.  Also the session 5 file except_exercise.py

students/douglas_klos/session4/mailroom2.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+def thank_you_menu():
+    """ Menu for Thank You options """
+
+    while True:
+        name_input = input(THANK_YOU_PROMPT)
+        if name_input.lower() in ('q', 'quit'):
+            return
+        elif name_input.lower() in('l', 'list'):
```
ldconejo: This _elif_ statement is not necessary, as it is preceded by a _return_, you can change it to an _if_.

students/douglas_klos/session4/trigrams.py
```diff
@@ -1,30 +1,44 @@
...
     outtab = ' '
     transtab = str.maketrans(intab, outtab)
 
-    # startline = '*** START OF THIS PROJECT GUTENBERG EBOOK'
-    startline = 'Produced by an anonymous Project Gutenberg'\
-                'volunteer and Jose Menendez'
-    # endline = '*** END OF THIS PROJECT GUTENBERG EBOOK'
-    endline = 'End of the Project Gutenberg EBook'
+    startline = '*** START OF THIS PROJECT GUTENBERG EBOOK'
+    endline = '*** END OF THIS PROJECT GUTENBERG EBOOK'
```
ldconejo: This works well, but it obviously limits the types of files you can run through your script (for example, _sherlock_small.txt_ would not work. One thing to consider in the particular case of trigrams is that effect of lines of text that are not part of the main novel gets reduced for larger file, as they become statistically less likely to show up as part of the generated text (because there are other combinations that are statistically more likely to show up).

It's very amusing to see how readable the output text is.

students/douglas_klos/session5/except_exercise.py
```diff
@@ -0,0 +1,59 @@
...
+# parameters. (pun intended)
+
+langs = ['java', 'c', 'python']
+
+try:
+    more_joke = more_fun(langs[0])
+except IndexError:
+    more_joke = more_fun(langs[1])
+else:
+    more_joke = more_fun(langs[2])
```
ldconejo: Consider here that _more_fun()_ does not return any values, so you're assigning _None_ to _more_joke_, which serves no purpose here. You can simply call _more_fun()_, without assigning it to a variable.

ldconejo:
Closing this pull request for now. I look forward to your submission for final review of the code shown here.

-Luis

PR #71, opened 2019-02-04
Submitting mailroom for feedback

UncleanlyCleric:
I had already used dictionary in my initial build... so I wasn't sure if you wanted me to do something different there.


ldconejo:
Hi Jesse,

I'm using this pull request to reply to your question on the trigram problem, as it's easier to display code. Look at the code changes for _parse_file()_ below:
```
def parse_file():
    translator = str.maketrans('', '', string.punctuation)
    with open('sherlock.txt') as input_file: 
        full_list = []
        for line in input_file:
            full_list.extend(line.translate(translator).split(' '))
    # At this point, we have a big list with all the words in the text, no trigrams yet

    sherlock_dict = {}
    for i in range(0, len(full_list) - 2):
        # This will make the trigram key a two-element tuple
        trigram_key = (full_list[i], full_list[i+1])
        trigram_value = full_list[i+2]
        # Check if this key is already in the dictionary
        if trigram_key in sherlock_dict:
            # Append it to the list of values associated to the key
            sherlock_dict[trigram_key].append(trigram_value)
        else:
            # Initialize the new key
            sherlock_dict[trigram_key] = [trigram_value]
```
This creates a dictionary in which each entry has this format:

_('was', 'a'):  ['loud', 'mews', 'lawyer', 'delicate', 'lovely', 'quarter', 'group', 'smart']_

With that, you can then select a random key from _sherlock_dict_ to start your text, then pick a random word from the value list of that random key and append it to the text, then use the last two words to select the next key. For example:

If _('was', 'a')_ is the random key you selected, your text will begin with:

"was a"

Now, you will select a word from the list associated with _('was', 'a')_, _['loud', 'mews', 'lawyer', 'delicate', 'lovely', 'quarter', 'group', 'smart']_. Let's say the word selected was "lovely", now your text will become:

"was a lovely"

Your next key will be the last two words of the text: "a lovely", so the next key will be _('a', 'lovely')_ and so on.

-Luis

students/jesse_miller/session04/mailroom.py
```diff
@@ -0,0 +1,194 @@
...
+    Hopefully, this makes directories and files on first run for the listed
+    donors.  Hopefully
+    """
+    path = os.getcwd()
+    for k in donors:
+        current_donor = ""
+        current_donor = k
+        donor_math = donors[current_donor]
+        donor_total = sum(donor_math)
+        donor_avg = len(donor_math)
```
mdgregor: You have this same summation of the donor stuff further up the code.  Might that be a good candidate for a function to avoid duplication?

PR #75, opened 2019-02-09
Shabani mailroom2 pull request

Rockwell70:
Hello, this is my pull request for mailroom2. Trigrams is not yet complete.


students/jeff_shabani/session04/file_exercise.py
```diff
@@ -0,0 +1,103 @@
...
+                                'typescript',
+                                'vb',
+                                'visualbasic',
+                                }
+        # 1st list containing text to right of ":" and stripped of white space
+        languages = []
+        # final list of acceptable languages cleaned of extra characters and white space
+        cleaned_languages = []
+
+        with open(file, 'rt') as f:
```
mdgregor: `file` is a python keyword, can't use it as a variable

students/jeff_shabani/session04/file_exercise.py
```diff
@@ -0,0 +1,103 @@
...
+
+def main():
+    if os.name == 'nt':
+        root = Path(r'D:/JRS/Python/UW/Intro_Class/students/jeff_shabani/session04')
+        src = Path(r'D:/JRS/Python/UW/Intro_Class/students/jeff_shabani/session04/files_to_copy')
+        destination = Path(r'D:/JRS/Python/UW/Intro_Class/students/jeff_shabani/session04/copied_files')
+    else:
+        root = Path('/Volumes/GASecure/JRS/Python/UW/Intro_Class/students/jeff_shabani/session04')
+        src = Path('/Volumes/GASecure/JRS/Python/UW/Intro_Class/students/jeff_shabani/session04/files_to_copy')
+        destination = ('/Volumes/GASecure/JRS/Python/UW/Intro_Class/students/jeff_shabani/session04/copied_files')
```
mdgregor: you don't need all this hardcoding of file names based on operating system.  you can do the following
```
root = Path(os.getcwd())
src = Path(f"{root}/files_to_copy")
destination = Path(f"{root}/copied_files")
```

mdgregor: not really sure you need the `Path()` call either

students/jeff_shabani/session04/trigrams.py
```diff
@@ -0,0 +1,53 @@
...
+
+word_reg = re.compile(r'\W+')  # regex to split text by white space
+
+
+# read in the text file and parse the words
+def trigram(file):
+    pair_dict = dict()
+    word_list = []
+
+    with open(file, 'rt') as infile:
```
mdgregor: `file` is a python keyword, need to use something else

mdgregor:
You need to remove your .idea/ stuff from your submission before I can merge.  Also add `.idea/` to your `.gitignore`  having just `.idea` won't do the trick

students/jeff_shabani/session04/file_exercise.py
Rockwell70: I removed the Path call from root and destination with no negative affects. When I remove it from src I get this error "AttributeError: 'str' object has no attribute 'iterdir'". I tried different fixes using the os module and nothing helped.

students/jeff_shabani/session04/file_exercise.py
Rockwell70: Fixed.

students/jeff_shabani/session04/trigrams.py
Rockwell70: Fixed

Rockwell70:
.idea stuff removed and .idea/ added to .gitignore

Rockwell70:
Thanks for the kind words on my mailroom2 code. It's a really interesting project and you get a sense of accomplishment when it works.

Rockwell70:
This is my trigrams submission. I think I've spent 3X the time I've spent on mailroom2 on this one. Cutting my losses at this point.

ldconejo:
Hi Jeff,

I made a few changes to your code for trigrams, moving the _populate_word_list()_ and _build_pair_dictionary()_ nested functions out of their nest and calling them from the main flow. I changed the dictionary keys back to tuples, as it turns out they're more convenient in that fashion (I preferred strings before).

I also created a new function, called _build_new_text()_, which takes _pair_dict_ as input and outputs the new text. I won't include the code of that one here, but the logic:
```
#!/usr/bin/env python3

import random
import re

word_reg = re.compile(r'\W+')  # regex to split text by white space

def populate_word_list(text_file):
    word_list = []
    with open(text_file, 'rt') as infile:
        lines = infile.readlines()
        for line in lines:
            for word in word_reg.split(line):
                if word:
                    word_list.append(word)
    return word_list

def build_pair_dictionary(word_list):
    pair_dict = {}
    for i in range(len(word_list) - 2):
        # Converted in back to a tuple
        pair = tuple(word_list[i:i + 2])
        follower = word_list[i + 2]
        if pair not in pair_dict:  # Check if the pair is already in the dictionary
            pair_dict[pair] = [follower]
        else:
            pair_dict[pair].append(follower)
    return pair_dict

def build_new_text(pair_dict):
    # selected a random key from pair_dict and assign it to trigram_key
    # add the two words in the key to full_text
    # while trigram_key exists in pair_dict, do the following:
    # 1. Get the word list associated to trigram_key in pair_dict
    # 2. Select a random word from that list.
    # 3. Add that word to full_text.
    # 4. Create a new trigram_key tuple, with the second element of
    #    trigram_key and the random word for the list
    # This will repeat itself until you find an key that doesn't exist
    # in pair_dict or reach a limit (for example, 1000 iterations)
    full_text = ''
    return pass

if __name__ == '__main__':
    word_list = populate_word_list('sherlock.txt')
    pair_dict = build_pair_dictionary(word_list)
    full_text = build_new_text(pair_dict)
    print(full_text)
```
Give this a try and submit a new pull request with updated code for trigrams.py, you're very close at this point.

-Luis

PR #76, opened 2019-02-09
Code Review on Lesson4\trigrams.py

awhitty1:
Hello Luis,

I was wondering if you could review my code (import random) is not producing the expected new text.
My output is not generating new text applied by the trigrams.


students/alex_whitty/lesson4/trigrams.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
+
+with open('double_take.txt', 'r') as f:
+
+    for line in f:
+        print(line, end='')
+
```
ldconejo: You should turn file operations into a function that returns a list of words. This code would do the trick:

```
def parse_file():
    '''
    Takes a hardcoded file name and returns a list of words
    '''
    # Setup a translator to remove punctuation
    translator = str.maketrans('', '', string.punctuation)
    # Setup an empty list to contain the full list of words in the text
    full_list = []
    with open('double_take.txt', 'r') as f:
        for line in f:
            # This will take each line, remove punctuation and then split it
            # into a list of words and attach this list to the full_list
            # note the use of "extend()" instead of "append()"
            # TODO: Figure out how to remove carriage returns
            full_list.extend(line.translate(translator).split(' '))
    return full_list
```

students/alex_whitty/lesson4/trigrams.py
```diff
@@ -0,0 +1,63 @@
...
+            trigrams[pair].append(follower)
+
+
+    return trigrams
+
+# build up the dict here!
+
+def build_dict(words):
+
+    import random
```
ldconejo: Don't do the import from inside of a function. This should be at the top of the file, along with other _import_ statements.

students/alex_whitty/lesson4/trigrams.py
```diff
@@ -0,0 +1,63 @@
...
+
+    random.choice(words)
+
+    new_text = 'from outer'
+
+    if words == {('from', 'outer'): ['Space']}:
+        new_text.append("Space")
+        return new_text
+
```
ldconejo: This one is trickier, you need to do two different random selections. You will first select a random key from the trigrams dictionary, those will be the first two words of the new text. After that, randomly select one word from the list associated the key, that will be the third word of the new text. Now, you need to take the second word of that first key and the word you picked from the list and form a tuple, which will be the next key. You need to repeat this process either until you create a key that is not in the dictionary or you meet some other stop condition of your own choosing. The code below starts the process:

```
def build_dict(words):
    starter = random.choice(list(words.keys()))
    print(starter)
    # Start the text with the random key
    # now turned into a string
    new_text = ' '.join(starter)

    # Now, pick a random word from the list
    # associated to the key
    key_value = words[starter]
    next_word = random.choice(key_value)

    # Add next_word to new_text
    new_text += ' ' + next_word
    print(new_text)

    # Form the next round, using the second
    # word of the starter and next_word
    # and repeat the process
```

ldconejo:
Close this pull request for now, as feedback for _trigrams.py_ has been provided.

-Luis

students/alex_whitty/lesson3/mailroom.py
```diff
@@ -0,0 +1,106 @@
...
+
+    answer = input("Please enter Full Name.>>>")
+    if answer.lower() == 'list':
+        for donor in donor_db.keys():
+            print(donor)
+    else:
+        if answer in donor_db:
+            current_donor(answer)
+        else:
+            add_donor()
```
ldconejo: This one should be _add_donor(answer)_, so that the name of the donor is passed to _add_donor()_, preventing the user from being asked for the name again (a bit confusing).

students/alex_whitty/lesson3/mailroom.py
```diff
@@ -0,0 +1,106 @@
...
+
+def add_donor():
+
+
+    new_donor = input("Name of the new donor to add?").title()
+    donor_amount = input("What is the donation amount?")
+    donor_db[new_donor] = [float(donor_amount)]
+    print(donor_db)
+    print('Thanks, {} for your generous donation.'.format(new_donor))
+
```
ldconejo: _add_donor()_ should become _add_donor(new_donor)_, so that you don't need to ask for the name of the donor again.

students/alex_whitty/lesson3/mailroom.py
```diff
@@ -0,0 +1,106 @@
...
+    r3 = '{:20}{:>12}{:>10}{:>18}'.format('Jeff Bezos', '$877.33', '1', '$877.33')
+    r4 = '{:20}{:>12}{:>10}{:>18}'.format('Paul Allen', '$708.42', '3', '$236.44')
+    r5 = '{:20}{:>12}{:>10}{:>18}'.format('Elon Musk', '$200.32', '1', '$200.32')
+
+    print(r1)
+    print(r2)
+    print(r3)
+    print(r4)
+    print(r5)
+
```
ldconejo: I missed _report()_ the first time around. Right now, it's hardcoded with a list of donors. This will not work, since you need to be able to add new donors and/or donations for existing donor, so the report needs to be dynamically created from _donor_db_.

ldconejo:
Hi Alex,

The mailroom from lesson 3 looks better now. I've added some comments on what needs to be fixed. The biggest part is changing the _report()_ function to get its information from _donor_db_, instead of hardcoding the names of the donors and the values of their donations. Thanks for putting the extra work to get these fixed.

-Luis

PR #77, opened 2019-02-09
Lesson 04 Assignment: Mailroom, Part 2

pauleclifton:


mdgregor:
```
>>> import time
>>> time.strftime("%b %d %Y %H:%M:%S")
'Feb 09 2019 12:12:38'
>>> 
```

students/paul_c/session04/mailroom.py
```diff
@@ -0,0 +1,145 @@
...
+def send_all():
+    """
+    Function to send a thank you letter to all donors.
+    This function sums total donations and writes letters to timestamped files.
+    """
+    for key, value in donors.items():
+       # I could not figure out how to get the timestamp to work with the filename for open(). I kept getting a file does not exist error.
+       # timestamp = datetime.now().strftime('%D')
+       # filename = "{}.txt.".format(key) + str(timestamp)
+        filename = "{}.txt".format(key)
```
mdgregor:
```
filename = f"{time.strftime("%b %d %Y %H:%M:%S")}-{key}.txt"
```

students/paul_c/session04/mailroom.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+    header = ("Donor Name", "| Total Given", "| Num Gifts", "| Average Gift")
+    row = " ".join(["{:20s} {:>20s} {:>20s} {:>20s}"]).format(*header)
+    header_length = len(row)
+    print("\n" + row)
+    print("=" * header_length)
+    for key, value in sorted(donors.items()):
+        value_sum = str(sum(value))
+        value_len = str(len(value))
+        value_ave = str(sum(value)/(len(value)))
```
mdgregor: you don't need to cast them as `str` before using in a string

students/paul_c/session04/mailroom.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+    for key, value in donors.items():
+       # I could not figure out how to get the timestamp to work with the filename for open(). I kept getting a file does not exist error.
+       # timestamp = datetime.now().strftime('%D')
+       # filename = "{}.txt.".format(key) + str(timestamp)
+        filename = "{}.txt".format(key)
+        file_content = '{greeting}' '{body}' '{closing}' '{signature}'.format(**email).format(key, sum(value))
+        files = open(filename, "w")
+        files.write(file_content)
+        print(str(files) + "  ---File created.\n")
```
mdgregor: make sure you call `files.close()` here if you aren't going to use the `with` command

mdgregor:
> I wasn't able to figure out how to add timestamps to the filename using open(). No matter how I formatted it, I would get a similar error to the following.
> 
> Enter here: Traceback (most recent call last):
> File "./mailroom.py", line 145, in
> main()
> File "./mailroom.py", line 137, in main
> main_menu.get(user_input)()
> File "./mailroom.py", line 33, in send_all
> files = open(filename, "w")
> FileNotFoundError: [Errno 2] No such file or directory: 'Anne Ant.txt.02/09/19'

Most likely the reason you got that error was because of the `/` and all the `.` in the file name.  If your on windows it really won't like this.  See my noted above on how best to insert that info.

For reference: https://www.tutorialspoint.com/python/time_strftime.htm

students/paul_c/session04/mailroom.py
pauleclifton: If I don't convert them to strings, I get exceptions like this.

Traceback (most recent call last):
  File "./mailroom.py", line 147, in <module>
    main()
  File "./mailroom.py", line 139, in main
    main_menu.get(user_input)()
  File "./mailroom.py", line 54, in create_report
    row_format = (key, "$" + value_sum, value_len, "$" + value_ave)
TypeError: can only concatenate str (not "float") to str

students/paul_c/session04/mailroom.py
pauleclifton: Added. Thanks!

students/paul_c/session04/mailroom.py
pauleclifton: Oh, this was a filesystem error, not a code error. Thanks for the tip!

PR #78, opened 2019-02-09
Updated Trigrams, really need feedback, this is a hard assignment for me.

UncleanlyCleric:


students/jesse_miller/session04/trigrams.py
```diff
@@ -16,55 +13,62 @@ def parse_file():
...
     Main run list
     """
-    parse_file()
+    try:
+        filename = sys.argv[1]
+    except IndexError:
+        print("You must pass in a filename")
+        sys.exit(1)
+
+    in_data = parse_file()
```
ldconejo: Change _parse_file()_ to take a filename parameter, so that you use the one passed on the arguments:
```
parse_file(filename)
```

students/jesse_miller/session04/trigrams.py
```diff
@@ -4,10 +4,7 @@
 """
 import string
 import random
-#import os
-#import re
-import itertools
-
+import sys
 
 def parse_file():
```
ldconejo: Modify _parse_file_ to take and use a filename, since you're asking the user for one :-)

students/jesse_miller/session04/trigrams.py
```diff
@@ -16,55 +13,62 @@ def parse_file():
     dictionary.
     Thanks for the help Luis!
     """
-    temp_dict = {}
-    sherlock_dict = {}
-    translator = str.maketrans('', '', string.punctuation)
-    sherlock_input = set(line.strip() for line in open('sherlock_small.txt'))
+    startline = 'Produced by an anonymous Project Gutenberg'\
+                'volunteer and Jose Menendez'
+    endline = 'End of the Project Gutenberg EBook'
```
ldconejo: Don't worry about _startline_ and _endline_ right now, just process the entire text.

students/jesse_miller/session04/trigrams.py
```diff
@@ -16,55 +13,62 @@ def parse_file():
...
+        start = 0
+        for line in input_file:
+            if line[:len(startline)] == startline:
+                start = 1
+                continue
+            if not start:
+                input_file = [line.rstrip('\n') for line in input_file]
+            if line[:len(endline)] == endline:
+                break
+    return full_list
```
ldconejo: Note that you're returning _full_list_ (an empty list), you could just read the file, remove carriage (which you're doing) and return _input_file_ for _make_words_ to work on it:

```
def parse_file(filename):
    with open(filename) as input_file:
        input_file = [line.rstrip('\n') for line in input_file]
    return input_file
```

students/jesse_miller/session04/trigrams.py
```diff
@@ -16,55 +13,62 @@ def parse_file():
...
+    except IndexError:
+        print("You must pass in a filename")
+        sys.exit(1)
+
+    in_data = parse_file()
+    words = make_words(input_file)
+    word_pairs = build_trigram(full_list)
+    new_text = build_text(sherlock_dict)
 
+    print(new_text)
```
ldconejo: I changed _main()_ to the code below, to match the other suggested changes. Note that it ends by displaying the keys (2-word tuples) and values (lists with 1+ elements) of the trigram dictionary.  The next step will be to add some code to _build_text()_. The logic for that is somewhat simple:

1. Pick a random key in the trigram dictionary.
2. Get the list of words associated to that key.
3. Pick a random word from the list of words.
4. Add the two words from the key and the word from the list to the new text.
5. Create a new 2-word tuple formed by the second word of the current key and the word for the list. This will be key for the next iteration.
6. Get the list of words associated to the new key.
7. Select a random word from that list and add it to the new text.
8. Repeat steps 5-7 until you create a key that doesn't exist in the trigrams dictionary or run into another stop condition you define.

```
def main():
    """
    Main run list
    """
    try:
        filename = sys.argv[1]
    except IndexError:
        print("You must pass in a filename")
        sys.exit(1)

    input_file = parse_file(filename)
    words = make_words(input_file)
    word_pairs = build_trigram(words)
    for key, value in word_pairs.items():
        print(f'Key: {key}, Value: {value}')
```

ldconejo:
Closing this pull request for now, while feedback provided is incorporated into the code.

UncleanlyCleric:
Okay, awesome.  Thank you!  I wish I knew why I was struggling so much with this one.

PR #79, opened 2019-02-10
Submit session04-trigrams

yiyunxu:


ldconejo:
Trigrams assignment already reviewed. Good job.

PR #80, opened 2019-02-10
Adding trigrams... 

UncleanlyCleric:


ldconejo:
Nice job! This is my favorite short story so far:

_once that the lust of the ceiling Round his brow so dark as it did not himself know The initials were of a number of constables with an ounce of shag tobacco and as it afterwards transpired the Countess deposed to having been there when I mentioned to you  It came right away with the same lines at Munich the year after the last time You said that she must feel to you I assure you that I was very drunk and when the police might not fall upon the window not long remain unavenged Why Watson let us!_

Aside from that, it was nice to see you code score 10 out of 10 on Pylint.

-Luis

UncleanlyCleric:
Awesome!  I’m trying to pay more attention to the PEP8 and linting where I can.  Thanks!

PR #81, opened 2019-02-10
Lesson 04 Assignment: Kata Fourteen: Tom Swift Under Milk Wood 

pauleclifton:
I keep having the thought that I missed something in the build_text function. Like, I left something out. Is there something else that function is supposed to do?

Currently, the function finds a random key, finds that value, appends the key and value to a new list, and cleans up the list a little with join.


ldconejo:
Hi Paul,

The output looks good (actually, quite readable for text generated with trigrams :-) ). There are some minor things to correct from a PEP-8 perspective, nothing that I would consider big enough to subtract points, but worth taking a look at. I'll comment on those directly on the code. Good job!

-Luis

students/paul_c/session04/kata_fourteen.py
```diff
@@ -0,0 +1,90 @@
...
+import sys
+import random
+
+"""
+This program was tested using the sherlock_small.txt and a_case_of_identity.txt files located in the working directory.
+
+"""
+
+def build_text(word_pairs):
+        """
```
ldconejo: You doubled the indentation within this function (8 spaces instead of 4). Not a big issue, but in general, the indentation should be consistent across your code.

PR #85, opened 2019-02-11
Resubmission of trigrams

Rockwell70:
Hello,  this is my resubmission of trigrams. Please review trigrams2.


Rockwell70:
Oops. Sorry about that. Definitely not intended. Do I need to do another pull?

mdgregor:
I'd try to pull from master into your repo.  If that doesn't work, then I might suggest a re-clone

Rockwell70:
I did both (pull and the re-clone) yesterday. I made a small change to trigrams2 in the master and started a pullt request but the problem is that it shows all the changes from yesterday (250+) and then the single change from today. Is there a way to only pull the change from today?

mdgregor:
Are you available now to do a zoom session?

mdgregor:
https://washington.zoom.us/j/3774146329

Rockwell70:
Let's see I might be blocked at work.

Sent from my BlackBerry - the most secure mobile device

PR #87, opened 2019-02-11
Week 4

dac2229:
made changes for lesson 3 and created lesson 4 files.


students/daniel_carrasco/session03/mailroom.py
```diff
@@ -0,0 +1,41 @@
...
+        row_format ="{:>15}" * 3
+        print(row_format.format('Name','Donation','Amount'))
+        for name,amount,time in zip(donors,cash,times):
+            print(f'{name.title():>15}{"$ "+str(amount):>15}{time:>15}')
+
+    response3 = input('Add a name?').lower()
+    if response3 == 'yes':
+        response4 = input('Add first and last').title()
+        donors.append(response4)
+        response5 = input('Donation amount?')
```
mdgregor: `response(x)` is not a good variable name.  It tells me nothing about what it is.  It should describe the value held within, like `donation_amount` or `first_and_last_name`

students/daniel_carrasco/session03/slice.py
```diff
@@ -0,0 +1,44 @@
+def firstandlast(value):
```
mdgregor: need `_` between words

students/daniel_carrasco/session04/mailroom.py
```diff
@@ -0,0 +1,65 @@
...
+    row_format ="{:>15}" * 4
+    print(row_format.format('Name','Donation ($)','Amount','Average ($)'))
+    for key,value in donors.items():
+        print(f'{key.title():>15}{value[0]:>15.2f}{value[1]:>15}{value[0]/value[1]:>15.2f}')
+    return None
+def letter():
+
+    for key,value in donors.items():
+        filename = open(key.title()+"_"+today+'.txt','w+')
+        filename.write(f'Dear {key.title()},\n\n\
```
mdgregor: you could use triple `'''` and not need to add all the `\n`s.  Or at least less of them :)

students/daniel_carrasco/session04/mailroom.py
```diff
@@ -0,0 +1,65 @@
...
+Thank you for your very kind donation of ${value[0]}.\n\n\
+It will be put to very good use.\n\
+\n\tSincerely,\n\
+\t\t-The Team')
+        filename.close()
+    return None
```
mdgregor: no need to return None

students/daniel_carrasco/session04/mailroom.py
```diff
@@ -0,0 +1,65 @@
...
+        if choice == 3:
+            main()
+
+    return choice
+def report():
+    row_format ="{:>15}" * 4
+    print(row_format.format('Name','Donation ($)','Amount','Average ($)'))
+    for key,value in donors.items():
+        print(f'{key.title():>15}{value[0]:>15.2f}{value[1]:>15}{value[0]/value[1]:>15.2f}')
+    return None
```
mdgregor: There needs to be two new lines between functions, also no need to `return None`

students/daniel_carrasco/session04/trigrams.py
```diff
@@ -0,0 +1,72 @@
...
+        lines.append(line.strip('\n'))
+    return lines
+
+def make_words(in_data):
+    new_data=[]
+    for line in in_data:
+        line = re.sub(r'[^\w\s]',',',line)
+        line = re.sub(r',,',' ',line)
+        line = re.sub(r',',' ',line)
+        line = re.sub(r'  ',' ',line)
```
mdgregor: I think it might be easier to use `.replace()` on the line to get rid of punctuation and spaces rather than regex

PR #90, opened 2019-02-11
Lesson4 submisson

woosukjeung:


students/WooseokJ/MailP2fixedWJ.py
```diff
@@ -0,0 +1,111 @@
...
+    print()
+    menu()
+#this prints out the database of the donors with their total amaounts and average amount donated
+def create_a_report():
+    sorted_donor_db = sorted(donor_db.items(), key=operator.itemgetter(1), reverse=True)
+    print("Printing report:")
+    title = ('Donor Name', 'Total Given', "Num Gifts", 'Average Gift')
+    print("{:<19} | {:^13} | {:^13} | {:^11}".format(*title))
+    print("-"*70)
+    for i, row in enumerate(donor_db):
```
ldconejo: Your code is currently crashing when _create_a_report()_ is called. One thing to change is to simply accessing _donor_db_. _donor_db_ is a dictionary, if you need to access the donors (dictionary keys) and the list of their donations (dictionary values), you just need to do:

for donor, donations in donor_db.items():
    # do something here to print maybe the name of the donor, 
    # the number of donations and the quantity of donations.
    # the name will be in the variable donor, you can do sum(donations)
    # for the total and len(donations) for the quantity.

students/WooseokJ/MailP2fixedWJ.py
```diff
@@ -0,0 +1,111 @@
...
+    while True:
+        fullname = input("Enter the full name of the donor typing list will give the names of current doners ")
+        if fullname.lower() == 'list':
+            for donor in donor_db:
+                print(donor)
+        elif fullname.lower() not in donor_list:
+            print(f'{fullname} does not exist in the donor data, '
+                  f'we will add {fullname} to the donor data.')
+            amount = float(input("Please enter the donation amount: "))
+            donor_db.append((fullname, [amount, 1]))
```
ldconejo: For this one, remember _donor_db_ is a dictionary, not a list, so it doesn't have an _append()_ method. You should be able to do _donor_db[fullname].append(...)_ since that would be calling the list of donations associated with a particular donor.

students/WooseokJ/TrigramWJ.py
```diff
@@ -0,0 +1,37 @@
...
+    random_text = []
+    while len(random_text) < len(pairs):
+        random_key = random.choice(list(pairs.keys()))
+        value = pairs.get(random_key)
+        random_text.append(list(random_key))
+        random_text.append(value)
+        new_list = [item for fixedlist in random_text for item in fixedlist]
+        clean_text = " ".join(new_list)
+#this is a promot that asks for the file that we want to scramble make sure to add .txt at the end of the file
+        return clean_text
```
ldconejo: Your code is working correctly up until _build_trigrams_dict()_. You only need to select a random key from the trigrams dictionary once, after that, you need to to continue the sequence. It should work like this:

1. Select a random key from _pairs_ (which is really the trigrams dictionary).
2. Add the two words in that random key to _random_text_.
3. Get the list of words associated to _random_key_ from the pairs dictionary.
4. Use random again to select one word from that list of words, let's call it _selected_word_.
5. Create the next key by taking the second word in the current key and _selected_word_.
6. Get the list of words associated with the new key.
7. Select a random word from that list.
8. Add the word to _random_text_.
9. Repeat steps 5 to 8 until you find a key that is not in the dictionary and/or until your _random_text_ reaches a certain number of words.

students/WooseokJ/TrigramWJ.py
```diff
@@ -0,0 +1,37 @@
...
+            trigrams[pair] = [follower]
+        else:
+            trigrams[pair] += [follower]
+    return trigrams
+#this mixes the words to create the new text
+def build_text(pairs):
+    random_text = []
+    while len(random_text) < len(pairs):
+        random_key = random.choice(list(pairs.keys()))
+        value = pairs.get(random_key)
```
ldconejo: An easier way to get the list of words associated to _random_key_ is to do _pairs[random_key]_.

ldconejo:
Please submit a new pull request once you have implemented the changes requested. Another suggestion is to add sub-folders for every lesson, so that your files for each lesson are clearly separated.

-Luis

PR #95, opened 2019-02-12
resubmitting for mailroom lesson 3 regrade.

dac2229:
I was asked to resubmit lesson 3 mailroom. I decided to make it different from lesson 4 mailroom since I didn't know about dictionary switching so I used if, elif, else for my menu. I kept the donors list as a dictionary since I knew how to use dictionaries before I started this class.

Thank you for letting me resubmit.

### DIFF
--- a/students/daniel_carrasco/session03/mailroom.py
+++ b/students/daniel_carrasco/session03/mailroom.py
@@ -1,23 +1,18 @@
 #!/usr/bin/env python3
-import tempfile
-from datetime import date
-import os
-
 def main():
-
-    print(tempfile.gettempdir())
-    terminate = ''
-    while terminate != 'quit':
-        print('\nChoose an action(1 - 3):\n\n\
-                1 - Send a Thank You to a single donor.\n\
-                2 - Create a Report.\n\
-                3 - Quit\n')
-        VALID_RESPONSES = (1,2,3)
-        choice = int(input(''))
-        while choice not in VALID_RESPONSES:
-            choice = int(input('Select 1 -3\n'))
-        arg_dict = {1:thankyou, 2:report,3:quit}
-        arg_dict[choice]()
+    print('\nChoose an action(1 - 3):\n\n\
+            1 - Send a Thank You to a single donor.\n\
+            2 - Create a Report.\n\
+            3 - Quit\n')
+    choice = int(input(''))
+    if choice == 1:
+        thankyou()
+    elif choice == 2:
+        report()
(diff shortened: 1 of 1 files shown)


students/daniel_carrasco/session03/mailroom.py
```diff
@@ -1,35 +1,52 @@
...
+    choice = int(input(''))
+    if choice == 1:
         thankyou()
-    pass
+    elif choice == 2:
+        report()
+    elif choice == 3:
+        quit()
+    else:
+        main()
```
mdgregor: calling main() inside of main won't really work well.  Also for choice 3, i'd suggest just doing sys.exit()

students/daniel_carrasco/session03/mailroom.py
```diff
@@ -1,35 +1,52 @@
 #!/usr/bin/env python3
 def main():
-
-    print('\noptions\n')
-    print('A = Send a Thank You')
-    print('\n')
-
-    response = input('what do you want to do?').upper()
-    if response == 'A':
+    print('\nChoose an action(1 - 3):\n\n\
```
mdgregor: There's not `while True` loop here which i think is why you were calling the `main()` in the else.  You should use the while true instead.

students/daniel_carrasco/session03/mailroom.py
```diff
@@ -1,35 +1,52 @@
...
+    return choice
+def report():
+    row_format ="{:>15}" * 4
+    print(row_format.format('Name','Donation ($)','Amount','Average ($)'))
+    for key,value in donors.items():
+        print(f'{key.title():>15}{value[0]:>15.2f}{value[1]:>15}{value[0]/value[1]:>15.2f}')
+    main()
+
+
+if __name__ == '__main__':
```
mdgregor: you have two `if __name__ ==  '__main__':` pieces.  This program won't work.

students/daniel_carrasco/session03/mailroom.py
```diff
@@ -1,35 +1,52 @@
...
+                2 - Enter Name.\n\
+                3 - Quit submenu\n'))
+        if choice == 1:
+            for keys in donors.keys():
+                print(keys)
+        if int(choice) == 2:
+            new_name = input('Enter full name\n').title()
+            donors.setdefault(new_name,[0,1])
+        if int(choice) == 3:
+            main()
```
mdgregor: this should just be `return` and not `main()`

PR #96, opened 2019-02-12
adding lesson04 for review

mmwheel:
I am resending this pull request.

### DIFF
--- a/students/marsha_w/lesson_04/mailroom2.py
+++ b/students/marsha_w/lesson_04/mailroom2.py
@@ -1,0 +1,146 @@
+#!/usr/bin/env python3
+
+'''Title: Mailroom exercise, part2
+Dev: Marsha Wheeler
+Date: 02/10/2019
+'''
+
+
+import datetime
+
+donor_dict = {"William Gates":[653772.32, 12.17],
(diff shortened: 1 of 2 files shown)


students/marsha_w/lesson_04/mailroom2.py
```diff
@@ -0,0 +1,146 @@
...
+def main():
+    user_response = input(user_prompt)
+
+
+    switch_func_dict = {
+    '1': show_list,
+    '2': send_thank_you,
+    '3': create_report,
+    '4': write_all,
+    '5': exit_program,
```
mdgregor: you want these values to be indented deeper than the var assignment.

students/marsha_w/lesson_04/trigrams.py
```diff
@@ -0,0 +1,79 @@
...
+Dev: Marsha Wheeler
+Date:02/09/2019
+'''
+
+import sys
+import random
+
+def clean_text(file):
+    good_words = []
+    with open(file) as f:
```
mdgregor: `file` is a python keyword, you'll need to use a different name.

PR #100, opened 2019-02-13
Lesson05 Assignments

sdlipp:
- Exception Exercise (along with Exception Test)

- Mailroom, Part 3.  Several Try/Except blocks were added but I didn't see a good place to use a list comprehension.  The few simple 'for' blocks in my code called the print function, which as explained in the text, is not an appropriate place to use a list comprehension.  Please let me know if you'd like me to re-arrange code to demonstrate use of list comprehensions.

### DIFF
--- a/students/ScottL/session05/Mailroom_Part3.py
+++ b/students/ScottL/session05/Mailroom_Part3.py
@@ -1,0 +1,141 @@
+#!/usr/bin/env python3
+
+import sys
+
+global_donation_list = {"Tom Petty": [5000, 15000],
+                        "Neil Diamond": [20000, 30000],
+                        "Elton John": [25000, 25000, 40000],
+                        "John Lennon": [25000],
+                        "Mick Jagger": [20000, 10000]}
+MENU_HEADER_1 = "\nMenu Options:"
+MENU_HEADER_2 = "\n" + "-" * len(MENU_HEADER_1)
+MENU_SELECTIONS = {1: "Send a 'Thank You' note to individual donor",
+                   2: "Display a report",
+                   3: "Send 'Thank You' letters to all donors",
+                   4: "Quit"}
+REPORT_HEADER_1 = "\n{:<26}{:<12}{:^17}{:>11}".format(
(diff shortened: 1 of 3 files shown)


students/ScottL/session05/Mailroom_Part3.py
```diff
@@ -0,0 +1,141 @@
...
+                    "Donor Name", "Total Dollars", "Number", "Avg Amount")
+REPORT_HEADER_2 = "\n" + "-" * len(REPORT_HEADER_1)
+DONOR_LIST_HEADER = "\nDonor Name:\n" + "-" * 20
+
+
+def display_user_menu():
+    """Display the contents of the menu options for the user."""
+    print(MENU_HEADER_1, MENU_HEADER_2)
+    for item in MENU_SELECTIONS:
+        print(item, ": ", MENU_SELECTIONS[item])
```
mdgregor: here is a good spot:
`print(item, ": ", MENU_SELECTIONS[item]) for item in MENU_SELECTIONS`

mdgregor: I'm not sure why in the text it says don't use them for print.  My only thought is that typically in comps they are used cause they are faster than a for loop.  But nothing in any of these programs really require speed, it more about practice and recognizing when they can be used.

students/ScottL/session05/Mailroom_Part3.py
```diff
@@ -0,0 +1,141 @@
...
+            input_donor_name()
+        else:
+            input_donation_amount(donor_name)
+
+
+def display_report():
+    """Call the function to calculate total & avg donations and then display a report."""
+    print(REPORT_HEADER_1, REPORT_HEADER_2)
+    for item in donation_calcs():
+        print("{:<26}${:>12,}{:^17}${:>11,}".format(item[0], item[1], item[2], item[3]))
```
mdgregor: Here is a good example of where not to use it as the print line is very long and it wouldn't make the comp very readable.

PR #101, opened 2019-02-14
Pull request for mailroom2.py

GFSUW:


students/ginafs/session04/mailroom2.py
```diff
@@ -0,0 +1,214 @@
...
+    print("1 - Thank a single donor")
+    print("2 - Create a Report")
+    print("3 - Send letters to all donors")
+    print("4 - Quit")
+    print()
+
+
+def add_donor(new_donor):
+    '''adds a new donor to the end of the list'''
+    DONATIONS.append(new_donor)
```
ldconejo: Note that _DONATIONS_ is a dictionary, so it has no _append()_ method.

students/ginafs/session04/mailroom2.py
```diff
@@ -0,0 +1,214 @@
...
+    amount = input(f"Enter new donor donation amount > ")
+    return amount
+
+
+def process_donation(new_donor_flag, donor_index):
+    '''process the donation. If it is a new donor, add donor and donation to donations dict.
+    if it is an existing donor, add donation to the list of donations'''
+    amount = get_donation_amount()
+    if new_donor_flag:
+        DONATIONS.append(list(int(amount)))
```
ldconejo: Same as on line 173, no append method on _DONATIONS_

PR #103, opened 2019-02-15
Session5, mailroom3, try/except

Douglas-Klos:
Try / Except statements for error trapping bad user input and when trying to write to directories without permission.

I'm not sure where I'd use comprehensions in this code.  Before using a dictionary there were a couple I had used, but removed them after adding dicts.  If there's a place that they could be used in this I'm open to suggestions but

### DIFF
--- a/students/douglas_klos/session5/mailroom3.py
+++ b/students/douglas_klos/session5/mailroom3.py
@@ -1,8 +1,8 @@
 #!/usr/bin/env python3
-""" Mailroom V3.0, now with more error trapping """
+""" Mailroom V4.0, now with more error trapping """
 
 # Douglas Klos
-# February 14th, 2019
+# February 15th, 2019
 # Python 210, Session 5, Mailroom v3
 # mailroom3.py
 
@@ -112,14 +112,14 @@ def add_donation():
     while True:
         try:
             donation_input = float(input('Please enter a donation: '))
+        except ValueError:
+            print(f'Donation must be a positive number')
+        else:
             if donation_input < 0:
                 print('Donation must be a positive number')
             else:
                 mailroom_db[name].append(donation_input)
                 break
-        except ValueError:
-            print(f'Donation must be a positive number')
-
 
 def remove_donor(name_input=''):
     """ Remove donor from the database """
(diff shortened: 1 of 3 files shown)


students/douglas_klos/session5/mailroom3.py
```diff
@@ -0,0 +1,300 @@
...
+                    '\t\t\tThe Team\n')
+
+
+def display_database():
+    """ Displays all donors and their donations """
+
+    print()
+
+    for key in mailroom_db:
+        print(f'{key:>24} : {mailroom_db[key]:}')
```
mdgregor: you could use a comp here.  I know the text says something about dont' use for prints, but yes, there isn't really a spot for this in this assignment except to recognize where they COULD be used.  Typically we only use comps for speed as they are faster than for loops.

students/douglas_klos/session5/mailroom3.py
```diff
@@ -0,0 +1,300 @@
...
+        try:
+            donation_input = float(input('Please enter a donation: '))
+        except ValueError:
+            print(f'Donation must be a positive number')
+        else:
+            if donation_input < 0:
+                print('Donation must be a positive number')
+            else:
+                mailroom_db[name].append(donation_input)
+                break
```
mdgregor: I think you could combine these into 1 while true loop

students/douglas_klos/session5/mailroom3.py
```diff
@@ -0,0 +1,300 @@
...
+            donation_input = float(
+                             input("Please enter donation name to remove: "))
+        except ValueError:
+            print(f'Invalid Entry')
+        else:
+            for donation in mailroom_db[name_input]:
+                if donation_input == donation:
+                    mailroom_db[name_input].remove(donation_input)
+                    return
+            print(f'Donation {donation_input} not {name_input} not found')
```
mdgregor: again these could be combined into one.

students/douglas_klos/session5/mailroom3.py
mdgregor: actually nevermind :)

PR #104, opened 2019-02-15
Submitting for a review

UncleanlyCleric:
I'm still having problems with the menu except entry.  It works until you type in something in VALID_INPUT.  Once that's done, ctrl-c will crash it... unless you type something wrong in the input prompt.  Then it cycles correctly again.

I'm think that it's not properly starting the while loop if you try to hit the except before reinitializing through input.  I have no idea how to fix it though.


ldconejo:
Hi Jesse,

Try the following instead of your nested _while_. I'm incorporating the _try/except_ block to the main _while True_, if one of the two valid exceptions occur, it will simply go to the next loop. If there is no exception, it will check if the answer was valid. If it's invalid, it will do the same as the exception and go to the next loop.

```
def new_main():
    while True:
        try:
            response = input(PROMPT)
        except (KeyboardInterrupt, EOFError):
            continue
        if response not in VALID_INPUT:
            print('ERROR: Invalid option')
            continue
        menu_choice[response]()
```

Regards,

Luis

UncleanlyCleric:
Ah, so the problem was that the interrupt was only referring back to the start of the nested loop and not through the whole thing, until I added input, which reset the entirety. 

Thanks!

PR #105, opened 2019-02-15
Submitting mailroom.py for Session 05

UncleanlyCleric:


mdgregor:
Can you pull upstream so these conflicts can be resolved.

students/jesse_miller/session05/mailroom.py
```diff
@@ -1,184 +1,204 @@
...
+    '''
+    This will be the donation report section
+    '''
+    summary = []
+    headers = ['Donor Name', 'Total Given', 'Times Donated', 'Average Gift']
+
+    print(f"\n{'-'*80}\n{{:17}} | {{:<19}} | {{:<15}} | {{:<19}}\n{'-'*80}"\
+    .format(headers[0], headers[1], headers[2], headers[3]))
+
+    for k_ey, v_alue in donors.items():
```
mdgregor: what is with the random `_`'s here?

UncleanlyCleric:
Oh, right.  I was trying to find something in the linter so I changed all the silly stuff it complained about, but forgot to change it back.  Fixed now.

PR #106, opened 2019-02-16
Completed fixes on Session03\mailroom.py

awhitty1:
Hello Luis,

I have completed the fixes on mailroom.py for review.

### DIFF
--- a/students/alex_whitty/session05/mailroom3.py
+++ b/students/alex_whitty/session05/mailroom3.py
@@ -1,0 +1,165 @@
+import sys  # imports go at the top of the file
+
+
+
+# Data Structure
+donor_db = {"William Gates, III": [653772.32, 12.17],
+            "Jeff Bezos": [877.33],
+            "Paul Allen": [663.23, 43.87, 1.32],
+            "Mark Zuckerberg": [1663.23, 4300.87, 10432.0],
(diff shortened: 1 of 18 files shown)


students/alex_whitty/session05/mailroom3.py
```diff
@@ -0,0 +1,165 @@
...
+        f"Sincerely,\n" \
+        f"The Fundraising Committee"
+    return letter
+
+''' Switch with the above functions "write_a_letter and write_a_letter_all" '''
+
+switch_func_dict = {
+    0: write_a_letter,
+    1: write_a_letter_all,
+}
```
mdgregor: This should be up at the top or bottom but not in among functions where it can get lost.

students/alex_whitty/session05/mailroom3.py
```diff
@@ -0,0 +1,165 @@
...
+        else:
+            print("Not a valid option!")
+
+
+
+if __name__ == "__main__":
+    # don't forget this block to guard against your code running automatically if this module is imported
+    main()
+
+sorted(donor_db, key=sort_key)
```
mdgregor: not sure what this is doing here.  It should be in the function that needs it or at the top of the file if you want it sorted on load.

PR #110, opened 2019-02-17
Submit Session05 exercise: Exceptions and mailroom part3

Mahmood08:


students/shirin_ak/Session05/mailroom.py
```diff
@@ -0,0 +1,165 @@
...
+#If the user types a name not in the list, add that name to the data structure and use it.
+#If the user types a name in the list, use it.
+#Once a name has been selected, prompt for a donation amount.
+#Convert the amount into a number 
+#Add that amount to the donation history of the selected user.
+#Finally, use string formatting to compose an email thanking the donor for their
+#generous donation. Print the email to the terminal and return to the original prompt.
+
+
+DONORS_NAME = {'william gates':[3, 3000.45],
```
mdgregor: This shouldn't be capitalized as its not a constant.

students/shirin_ak/Session05/mailroom.py
```diff
@@ -0,0 +1,165 @@
...
+    while True:       
+        donation = input("\nPlease Enter Valid Amount."
+                         "\n or Enter 'quit' to return to main menu." "\n$")        
+        if donation == 'quit':
+            return donation        
+        try:
+            donation = int(donation)
+            return float(donation)
+        except ValueError:
+             print("\nError.It is not valid amount.")
```
mdgregor: This should be reorganized a bit.  Also testing if a thing can be cast to int, but then casting as float doesn't quite make sense
```
        try:
            donation = input("\nPlease Enter Valid Amount."
                         "\n or Enter 'quit' to return to main menu." "\n$")        
            if donation.lower() == 'quit':
                return donation        
            return float(donation)
        except ValueError:
             print("\nError.It is not valid amount.")
```

PR #111, opened 2019-02-17
Trigrams submission

mjchang14:
I ran into a real problem getting the new trigram to loop through. I feel like I'm close, but I just don't have the time and energy right now to continue to beat myself up over it. I want to return to it at a later time though, and am open to any and all suggestions on how to think about this in a different way to make the loop work.

Thanks!


students/mjchang/session04/trigrams.py
```diff
@@ -0,0 +1,73 @@
...
+        for row in book:   
+            row = row.strip() #removes carriage lines
+            if row.startswith("End of the Project Gutenberg EBook"):
+                break    
+    #second part of removing punctuation, combine with extend
+            text.extend(row.translate(translator).split(' ')) #use extend to add multiple words to list
+    return text            
+       
+      
+    for word in text: #capitalize the word "I"
```
ldconejo: Because of the return statement that precedes it, this _for_ loop becomes unreachable.

ldconejo:
Hi Marissa, 

Implementation looks good. Whenever you have some time, run _Pylint_ or another linter tool to remove a few issues with how the code is written, the biggest one being a block of unreachable code. Other than that, your implementation is good.

Regards,

Luis

PR #112, opened 2019-02-17
submit session05 homework

yiyunxu:


students/elaine_x/session05/mailroom3.py
```diff
@@ -0,0 +1,130 @@
...
+        elif fullname.lower() not in donor_list:
+            print(f'{fullname} does not exist in current donor data, '
+                  f'adding {fullname} to the donor data.')
+            #prompt for a donation amount
+            while True:
+                try:
+                    donation_amount = float(input("Enter the donation amount: "))
+                    break
+                except ValueError:
+                    print("Donation amount has to be a number, try again.")
```
mdgregor: Need a `continue` here?

students/elaine_x/session05/mailroom3.py
```diff
@@ -0,0 +1,130 @@
...
+            #print(donor_db) #for testing
+            break
+        elif fullname.lower() in donor_list:
+            #prompt for a donation amount
+            while True:
+                try:
+                    donation_amount = float(input("Enter the donation amount: "))
+                    break
+                except ValueError:
+                    print("Donation amount has to be a number, try again.")
```
mdgregor: again with `continue`?

students/elaine_x/session05/mailroom3.py
yiyunxu: is the 'continue' necessary? I think the 'while' loop will keep running till user inputs something valid.

PR #113, opened 2019-02-17
Need code reviewed.

awhitty1:
Hello Luis/Michael,

I'm not sure about the comprehension coding, I created two of them in my mailroom3 file(getting the keys in donor_db). Are the comprehensions use to replace for loops within a function? Or, are they meant to replace the function with compact coding? Thanks for your assistance.

### DIFF
--- a/students/alex_whitty/session05/mailroom3.py
+++ b/students/alex_whitty/session05/mailroom3.py
@@ -10,9 +10,11 @@ donor_db = {"William Gates, III": [653772.32, 12.17],
             "Elon Musk": [2263.23, 3300.87, 15432.0],
             }
 
-dict_comprehension = \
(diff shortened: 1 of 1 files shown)


awhitty1:
The mailroom3.py is in my Session05 folder.

mdgregor:
Comps do replace _simple_ `for` loops, but not complex ones.  The advantage of comps is that they are executed faster than your traditional for loop.  You need a lot of iterations (items in the iterator) to see the difference though.  There are cases though when you might be iterating through thousands of record referencing from a database and you need to do it fast, so comps make a whole lot of sense.  What you did here is fine, the idea is just to get a feel for how they work and identifying where they should be used.

PR #114, opened 2019-02-17
added lesson05 files

GFSUW:


students/ginafs/session05/mailroom3.py
```diff
@@ -0,0 +1,222 @@
+''' Mailroom Part2 '''
+
+import datetime
+
+EDGE_STR = "| "
+DONATIONS = {
```
mdgregor: this shouldn't be capitalized as its not a constant.

students/ginafs/session05/mailroom3.py
```diff
@@ -0,0 +1,222 @@
...
+        donor_name = input("Enter a name from this list or the full name of the new donor : ")
+    except KeyboardInterrupt:
+        print_goodbye()
+        return False
+
+    #check if the donor is in the database or is a new donor
+    #if it is a new donor set the new_donor flag to true
+    new_donor = False
+    if donor_name not in DONATIONS:
+        new_donor = True
```
mdgregor: I think you don't need this block on lines 53-55.  See below

students/ginafs/session05/mailroom3.py
```diff
@@ -0,0 +1,222 @@
...
+        except KeyboardInterrupt:
+            print_goodbye()
+            return False
+        except ValueError:
+            print("Please enter a valid number for the donation amount")
+            continue
+        else:
+            break
+
+    if new_donor:
```
mdgregor: change this to `if donor_name not in DONATIONS:`

students/ginafs/session05/mailroom3.py
```diff
@@ -0,0 +1,222 @@
...
+        line_str += "{:25}".format(key) + EDGE_STR
+        # Total Given
+        sum_gifts = sum(DONATIONS[key])
+        line_str += "{: 14.2f}".format(sum_gifts) + EDGE_STR
+        # Num Gifts
+        num_gifts = len(DONATIONS[key])
+        line_str += "{: 12d}".format(num_gifts) + EDGE_STR
+        # Average
+        avg = sum_gifts/num_gifts
+        line_str += "{: 14.2f}".format(avg) + EDGE_STR
```
mdgregor: The comments strings you have here don't add anything as they are just copies of the variable names.  What your doing is simple enough to read that you probably don't need them.

students/ginafs/session05/mailroom3.py
```diff
@@ -0,0 +1,222 @@
...
+            out_tuple = (key, int(num_donations), float(total_amount))
+        else:
+            #donor with single donation does not need number of donations
+            out_str = SINGLE_DONATION_LETTER
+            out_tuple = (key, float(total_amount))
+
+        #create file and write into it
+        with open(file_name, 'w') as outfile:
+            outfile.write(out_str.format(*out_tuple))
+            outfile.close()
```
mdgregor: since your using the context manager of `with` you don't need `outfile.close()`

students/ginafs/session05/mailroom3.py
```diff
@@ -0,0 +1,222 @@
...
+
+        #Build the string to write into the file. Use tuples to send in values
+        out_str = ""
+        if num_donations > 1:
+            out_str = MULTIPLE_DONATION_LETTER
+            out_tuple = (key, int(num_donations), float(total_amount))
+        else:
+            #donor with single donation does not need number of donations
+            out_str = SINGLE_DONATION_LETTER
+            out_tuple = (key, float(total_amount))
```
mdgregor: variable names call `out_str` and `out_tuple` don't describe what the data is in those, just a type.

PR #115, opened 2019-02-17
Session 6, mailroom4, test driven development and refactoring, looking for feedback.

Douglas-Klos:
I've done quite a bit of refactoring to separate user input  from other functions and created a pytest test_mailroom4 file for mailroom4.

I've never done this sort of test driven development.  Was hoping for some feedback to know if I'm going about this correctly.


students/douglas_klos/session5/mailroom3.py
```diff
@@ -60,8 +60,15 @@ def display_database():
...
-        print(f'{key:>24} : {mailroom_db[key]:}')
+    # for key in mailroom_db:
+    #     print(f'{key:>24} : {mailroom_db[key]:}')
+
+    # Assignment asked for a comprehension.  Gregor suggested trying here.
+    # The text does state you wouldn't typcially use them for printing.
+    # IDK, this feels clunky to me, just mushing two lines of code into one.
+    # Though perhaps it's the pythonic way to do things.
+    # Included for the sake of completeness.
+    [print(f'{key:>24} : {mailroom_db[key]:}') for key in mailroom_db]
```
mdgregor: Yeah for prints you wouldn't, but comps are much faster than traditional for loops, but they are only for simple for loops or simple tested for loops.  You never want a comp to go more than 1 nested for loop otherwise they start to become unreadable.
```
for key, value in thing.item():
    for item in value:
        do_something(item)
```
^^^ example of 1 nested `for` loop.

Douglas-Klos: Yeah, makes sense.  In the first mailroom I had a 
`elif not any(name in sub_list for sub_list in mailroom_db):`
that seemed to work well.

Do you have any input on my test_mailroom4 and the refactoring on mailroom4?

Thanks :)

Douglas-Klos:
I've created pytest and unittest files for mailroom4, and unittest for args_kwargs_lab.  This is pretty foreign to me, want to know if I'm going about this correctly.

Thanks.

PR #116, opened 2019-02-17
Add try blocks to mailroom

pauleclifton:
Try blocks were added to the main and add_donation functions to prompt users to try again if they input bad data. I decided not to add a try block to the send_thankyou function because any unknown data would be received as a new donor name.

I had a hard time figuring out if I needed to use any comprehensions. The only time I generate a sequence from another sequence is in the send_all and create_report functions. In these functions a string is generated from iterating though a dictionary. I tried some things but I couldn't really figure out how to do a string comprehension out of a dictionary.

Please let me know if I missed the point of the assignment. I'll go back and add comprehensions if desired.

### DIFF
--- a/students/paul_c/session05/mailroom.py
+++ b/students/paul_c/session05/mailroom.py
@@ -1,0 +1,156 @@
+#!/usr/local/bin/python3
+
+import sys
+import time
+
+donors = {
+    "Anne Ant": [1.00],
+    "Bonnie Bug": [20.00, 10.00],
+    "Chuck Cat": [2.00, 2.00, 2.00],
+    "Donna Dog": [5000.00, 2500.00, 1.50],
+    "Edna Ent": [.30, .50, .10]
+}
+
+email = {
+    "greeting": "\nHello {}\n\n",
+    "body": "We would like to thank you for your generous donation of ${}.\n\n",
+    "closing": "Best Regards,\n",
+    "signature": "The Foundation\n\n"
+}
+
+
+def send_all():
+    """
+    Function to send a thank you letter to all donors.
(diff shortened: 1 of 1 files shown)


students/paul_c/session05/mailroom.py
```diff
@@ -0,0 +1,156 @@
...
+    """
+    This function prints out a report with the following parameters:
+    Donor Name, Total Given, Number of Gifts, Average Gift Amount
+    """
+    header = ("Donor Name", "| Total Given", "| Num Gifts", "| Average Gift")
+    row = " ".join(["{:20s} {:>20s} {:>20s} {:>20s}"]).format(*header)
+    header_length = len(row)
+    print("\n" + row)
+    print("=" * header_length)
+    for key, value in sorted(donors.items()):
```
ldconejo: Try converting this one to use list comprehension. To make it a bit simpler to read, I'd use f-strings. Here's a hint:

```
    result = [f"{key:20s} ... for key, value in ...)]
    for donor in result:
        print(donor)


```-Luis

ldconejo:
Hi Paul,

I added it a hint on how you can use list comprehension. One of the goals of the assignment is to have you experiment with comprehensions as an option. It this case, it probably won't improve performance, but it's still important to be familiar with how to use them.

-Luis

students/paul_c/session05/mailroom.py
pauleclifton: Thank you for the hint. I was able to turn that portion of code into the following.

donor_row = [f"{key:20s} {sum(value):>20.2f} {len(value):>20.2f} {sum(value)/len(value):>20.2f}" for key, value in sorted(donors.items())]

PR #117, opened 2019-02-18
review for mailp2 lesson 4 and lesson 5 submission 

woosukjeung:
should have addressed all the errors


students/WooseokJ/MailP2fixedWJFixed.py
```diff
@@ -0,0 +1,121 @@
...
+# In[ ]:
+
+
+import sys
+import datetime
+import operator
+
+
+def menu():
+    print(''''Menu
```
ldconejo: Because of how you're calling this function from _main()_, you should change this print to _return_, otherwise, _None_ will be printed each time you call _menu_. This happens because you're doing this:

`_input(menu())_ # input tries to print the output of menu, but menu() has no return statement, so it returns "None" by default.`

students/WooseokJ/MailP2fixedWJFixed.py
```diff
@@ -0,0 +1,121 @@
...
+def create_a_report():
+    print("Printing report:")
+    title = ("Donor Name", "| Total Given", "| Num Gifts", "| Average Gift")
+    row = " ".join(["{:<19} | {:^13} | {:^13} | {:^11}"]).format(*title)
+    length = len(row)
+    print("\n" + row)
+    print("=" * length)
+    for key, value in sorted(donor_db.items()):
+        given = str(sum(value))
+        gift = str(len(value))
```
ldconejo: Note that this one will always return "2", as that is the number of elements in the _value_ list associated to each donor. Since you're try to get the number of donations, you could do the following:

gift = value[1] # No need to convert to _str_, Python will do that for you when adding the to _row_format_

ldconejo:
Hi Wooseok,

Try to incorporate the changes suggest for lesson 4 into this submission as well. Also, whenever a new donation is entered, the value reflected is the value entered + 1, you'll need to fix that one as well.

_try/except_ blocks are well implemented. The assignment also requires you to use comprehensions, a good place could be the _create_a_report_ function.

-Luis

PR #118, opened 2019-02-18
adding_exception_and_mailroom3.py_scripts

mmwheel:
Ready for review -- Thanks!

### DIFF
--- a/students/marsha_w/lesson_05/mailroom3.py
+++ b/students/marsha_w/lesson_05/mailroom3.py
@@ -1,0 +1,139 @@
+#!/usr/bin/env python3
+
+'''Title: Mailroom exercise, part3
+Dev: Marsha Wheeler
+Date: 02/17/2019
+'''
+
+import sys
+import datetime
+
+
+donor_dict = {"William Gates": [653772.32, 12.17],
+              "Jeff Bezos": [877.33],
+              "Paul Allen": [663.23, 43.87, 1.32],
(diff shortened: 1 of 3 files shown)


students/marsha_w/lesson_05/mailroom3.py
```diff
@@ -0,0 +1,139 @@
...
+user_prompt = "\n".join(("Please choose from the options below:",
+                         "1 = To view a list of donors type 'list'",
+                         "2 - Send a Thank You to a single donor",
+                         "3 - Create a Report",
+                         "4 - Write a Thank you note to all donors",
+                         "5 - Exit the Program",
+                         ">>> "))
+
+# create a list of donor names
+donors = list(donor_dict.keys())
```
ldconejo: This creates a bug in your code when a new donor is added. Since _donors_ is not getting updated when a new donor comes in, every time you try to add a new donation to that new donor you code will think the donor is new and overwrite, instead of append, the new donation. 
 
You really don't need to use donors. If you want to check if a donor exists as a key in _donor_dict_ you can simply write:

```
if donor in donor_dict:
    # do something here

```
Consider removing check _donor_dict_ directly on line 43.

ldconejo:
Hi Marsha,

See my comment on your mailroom code for a small fix to correct a bug. Please submit a new pull request once that one is fixed.

Regards,

Luis

PR #119, opened 2019-02-18
lesson 5 assignment

dac2229:
Week 5 assignment. I think this went pretty well. I couldn't add any list comprehension to my mailroom but I can change my code to make list. I do use for statements but that's to print not to make a list.

The exceptions exercise has really helpful, I never knew about this!

### DIFF
--- a/students/daniel_carrasco/session05/mailroom.py
+++ b/students/daniel_carrasco/session05/mailroom.py
@@ -16,7 +16,13 @@ def main():
                     4 - Add new donor.\n\
                     5 - Add new donation.\n\
                     6 - Quit\n'))
-            arg_dict = {1: thankyou, 2: report, 3: letter, 4: addnewdonor, 5:addnewdonation, 6: quit}
+            arg_dict = {
+                1: thankyou,
+                2: report,
+                3: letter,
+                4: addnewdonor,
+                5: addnewdonation,
+                6: quit}
             if arg_dict[user_choice] == 'quit':
                 sys.exit()
             else:
@@ -87,11 +93,13 @@ def addnewdonor(user_input_name=' '):
                     input("Enter Donation Amount\n"))
         except ValueError:
             print("Input must be an integer, try again.")
-    return user_choice_name,user_choice_donation
+    return user_choice_name, user_choice_donation
+
 
 def addnewdonation():
     report()
-    user_input_name=input("\nEnter the name of donor adding a new donation.\nSee report for list of current donors\n\n")
+    user_input_name = input(
+        "\nEnter the name of donor adding a new donation.\nSee report for list of current donors\n\n")
     while True:
         keys = [key for key in donors]
         if user_input_name.title() in keys:
@@ -99,17 +107,12 @@ def addnewdonation():
                 user_choice_donation
             except NameError:
                 user_choice_donation = float(input("Enter Donation Amount\n"))
-                donors[user_input_name.title()][0]=donors[user_input_name.title()][0]+user_choice_donation
-                donors[user_input_name.title()][1]=donors[user_input_name.title()][1]+1
+                donors[user_input_name.title()][0] = donors[user_input_name.title()][0] + user_choice_donation
+                donors[user_input_name.title()][1] = donors[user_input_name.title()][1] + 1
             break
         else:
-            user_input_name,user_choice_donation = addnewdonor(user_input_name)
-
-
-
-
-
-
+            user_input_name, user_choice_donation = addnewdonor(
+                user_input_name)
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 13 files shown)


students/daniel_carrasco/session05/mailroom.py
```diff
@@ -0,0 +1,67 @@
...
+            if choice == 3:
+                main()
+        except ValueError:
+            print("Input must be an integer, try again.")
+            continue
+
+def report():
+    row_format ="{:>15}" * 4
+    print(row_format.format('Name','Donation ($)','Amount','Average ($)'))
+    for key,value in donors.items():
```
ldconejo: Use this one to implement comprehensions, it should be a relatively easy fit for the example provided as part of the exercise:

```
[print(donor) for donor in donors]
```

students/daniel_carrasco/session05/mailroom.py
```diff
@@ -0,0 +1,67 @@
...
+                sys.exit()
+            else:
+                arg_dict[user_choice]()
+        except ValueError:
+            print("Input must be an integer, try again.")
+        except KeyError:
+            print('Choice must be a menu input (1-4)')
+        continue
+
+def thankyou():
```
ldconejo: Try modifying the code so that new donations, as opposed to only new donors, can be entered.

students/daniel_carrasco/session05/mailroom.py
```diff
@@ -0,0 +1,67 @@
...
+    while True:
+        try:
+            choice = int(input('\nChoose an action(1-3):\n\
+                    1 - See Donor List.\n\
+                    2 - Enter Name.\n\
+                    3 - Quit submenu\n'))
+            if choice == 1:
+                for keys in donors.keys():
```
ldconejo: You can simply iterate the dictionary on this one, instead of calling _keys_:

```
for keys in donors:
```

ldconejo:
Hi Daniel,

Couple of changes to make:
1. I commented on a part of your code where you could try adding comprehension, which is one of the requirements of this assignment.
2. In general for your code, try to always add a space whenever you use a comma (for example, when entering multiple parameters into a function). It improves readability.

Submit a new pull request once you have implemented these and the suggestions added to the code. Very good use of exceptions, they add a lot of value in terms of making your code more robust.

-Luis

students/daniel_carrasco/session05/mailroom.py
ldconejo: One more comment on this one: If you run Pylint, it will complain that the _print()_ statement doesn't have a return, so your list comprehension will not really return a list. It will still work, but Pylint will make you feel bad about it :-)

One alternative is to take out the _print()_ statement a capture the output into a list for later printing. For example:

```
user_data = {
    "user01" : {
        "name" : "Robert",
        "last_name" : "Carnavaro"
    },
    "user02" : {
        "name" : "Adam",
        "last_name" : "Gouda"
    },
    "user03" : {
        "name" : "Luisa",
        "last_name" : "Ramirez"
    },
}

user_strings = [f'For user {user_name} name is {value["name"]} and last name {value["last_name"]}' for user_name, value in user_data.items()]

for user_string in user_strings:
    print(user_string)
```

dac2229:
I can't make a new pull request, can you see the changes I made?

dac2229:
Made changes to Week 5 as requested and added Week 6 assignment to this pull request.

Added as many test as I could, can you provide suggestions on how I could test my last two functions in mailroom (addnewname, andnewdonation). Also can you suggest how to improve my test.

Thank you!

ldconejo:
Hi Daniel,

Your unit tests do work, but the percentage of your mailroom code that is actually covered is pretty low. The assignment asks for _full site_ testing, which would imply 100% coverage, which is probably not realistic. A more reasonable number is 90% coverage.

One way to test coverage is using the _coverage_ module on Python. You can launch a coverage evaluation using this command:

```
$ python -m coverage run --include=mailroom.py -m pytest
```

After that, you can get a coverage report with this command (from the same folder where you ran you coverage evaluation):

```
$ python3 -m coverage report
```

In your case, the report is returning this:

```
Name          Stmts   Miss  Cover
---------------------------------
mailroom.py      96     68    29%
```

Coverage is at 29%, see if you can add some additional tests to get to or higher than 90%.

Additional to that, try adding docstrings to describe what the different functions in your code do, for example:

```
def test_get_report():
    '''
    Implements testing for the get_report() function
    '''
```
Regards,

Luis

dac2229:
Luis,
So for the functions where I have multiple user input statements, do I just
predefine the user inputs in the test function?few of my functions have
this so I excluded them since I was unsure of how to test them.
 Ex:
def print_ui ():
    ui = input('type anything')
    return ui
_________
def test_print_ui ():
    ui = 'test string'
    assert print_ui() = ui #would this work???

Thanks,
Daniel C.

ldconejo:
Hi Daniel,

Yes, in those cases, you could try using the _Mock_ module's patch decorator to override the _input_ built-in method. From one example that came up in the _Discussion_ section this week:

```
import mock
import unittest
import your_mailroom

class SampleTest(unittest.TestCase):
    @mock.patch('your_mailroom.input', mock.Mock(return_value = '54'))
    def test_correct_input(self):
        self.assertEqual(jeff_coyour_mailroomde.get_value('Enter a value:', int), 54)
```

In this case, _input_ is modified, *only* for the test being decorated with @mock.patch to always return the string '54'.

-Luis

PR #122, opened 2019-02-19
Submitting lesson 05 exercises

jroam142:


students/jeremy_m/lesson05_exercises/mailroom_part_3.py
```diff
@@ -0,0 +1,147 @@
...
+            print(donor)
+
+        thank_you_input = get_user_input('\n\n' + messages['thanks'])
+    
+    # Keep asking for donation_amount until user enters a number.
+    while True:
+        donation_amount = input(
+                'How much did {} contribute?\n---->'.format(thank_you_input))
+        try:
+            int(donation_amount)
```
mdgregor: this doesn't actually change the value of donation_amount to an `int`, its just testing if it can be an int.

students/jeremy_m/lesson05_exercises/mailroom_part_3.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def letters_for_all():
+    """ This function creates a directory in the cwd and fills that with an
+        individual file (letter) for each donor.
+    """
+
+    clear_screen()
+    print('Note: This operation will create a directory in the current working directory')
+    if get_user_input(messages['letters_confirmation']).lower() == 'y':
```
mdgregor: this function `get_user_input` doesn't really buy you anything as a function as its still just one line to do the same thing:
```
if input(messages['letters_confirmation']).lower() == 'y':
```

PR #126, opened 2019-02-20
mailroom_part4

Mahmood08:
Hi Luis,
  I started working on the Session 06 assignment. Can you please check if my approach is right?
Thank you .

### DIFF
--- a/students/shirin_ak/Session06/test_mailroom.py
+++ b/students/shirin_ak/Session06/test_mailroom.py
@@ -1,0 +1,18 @@
+import pytest
+from mailroom import create_report,send_thank_you_letter,add_donor
+
+
+
+def test_send_thank_you_letter():    
+    expected = "David,Thank you for your most recent donation of $2500.17."
+    assert send_thank_you_letter("David", 2500.17)== expected 
+
+
+
+
+def test_display_report():
+    report = create_report() 
+    assert create_report()== report
+
+
+print("passed")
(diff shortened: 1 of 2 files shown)


students/shirin_ak/Session06/test_mailroom.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+from mailroom import create_report,send_thank_you_letter,add_donor
+
+
+
+def test_send_thank_you_letter():    
+    expected = "David,Thank you for your most recent donation of $2500.17."
+    assert send_thank_you_letter("David", 2500.17)== expected
```
ldconejo: This one is good.

students/shirin_ak/Session06/test_mailroom.py
```diff
@@ -0,0 +1,18 @@
...
+def test_send_thank_you_letter():    
+    expected = "David,Thank you for your most recent donation of $2500.17."
+    assert send_thank_you_letter("David", 2500.17)== expected 
+
+
+
+
+def test_display_report():
+    report = create_report() 
+    assert create_report()== report
```
ldconejo: For this one, _report_ should be assigned the expected string that _create_report()_ will output, the way it is currently code, there's no way it would fail, as _report_ and _create_report()_ reference each other.

ldconejo:
Hi Shirin,

It looks good, see my comments on your tests.

Happy coding,

Luis

PR #127, opened 2019-02-22
submit session06 hw

yiyunxu:


students/elaine_x/session06/mailroom4.py
```diff
@@ -0,0 +1,165 @@
...
+#Session 06 - Mailroom Part 4
+#Elaine Xu
+#Feb 20, 2019
+###########################
+'''
+import sys
+#import datetime
+#import operator
+
+"""modifying this function for unit test
```
mdgregor: not sure why you are doing giant comment out blocks, just remove the unused code.

PR #128, opened 2019-02-22
Mailroom3 submission

mjchang14:
The exceptions I found were input related. I'm not sure if there's a better way to find them other than running the program multiple times and seeing what breaks, but I'm assuming that will come in a later lesson.


students/mjchang/session05/mailroom3.py
```diff
@@ -0,0 +1,118 @@
...
+                continue    
+            donor_db.setdefault(name, []).append(donation)
+            break
+        
+        elif name in donor_db.keys(): #adding donation for existing donor
+            name = name.title() #keep capitalization same as keys
+            #adding exception for scenario where input isn't a number
+            try:
+                donation = input("Enter the new donation amount: ")
+                donation = float(donation)
```
mdgregor: `donation = float(input("Enter the new donation amount: ")`

PR #133, opened 2019-02-24
Need code review

awhitty1:
Hello Luis/Michael, 

Having issues testing my code(mailroom4, test_mailroom4) seems to me refactor produces more errors when unit testing while using the unittest framework. Should I be able to test the code as is successfully before refactoring or do I have it backward? 
Thanks,
Alex

### DIFF
--- a/students/alex_whitty/session06/mailroom4.py
+++ b/students/alex_whitty/session06/mailroom4.py
@@ -1,0 +1,171 @@
+import sys  # imports go at the top of the file
+
+
+
+# Data Structure
+donor_db = {"William Gates, III": [653772.32, 12.17],
+            "Jeff Bezos": [877.33],
+            "Paul Allen": [663.23, 43.87, 1.32],
+            "Mark Zuckerberg": [1663.23, 4300.87, 10432.0],
+            "Elon Musk": [2263.23, 3300.87, 15432.0],
+            }
+
+
+
+''' First Comprehesion'''
+new_donor_dic = \
+     {d: donors for d, donors in donor_db.items()}
+
+# Input/Output
+prompt = "\n".join(("Mail Distribution Center!",
+          "Please choose from below options:",
+          "1 - Send a Thank You to a single donor",
+          "2 - Create a Report",
+          "3 - Send letters to all donors",
+          "4 - Exit",
+          ">>> "))
+
+# Features
+''' Second Comprehension'''
+dict_comp = {d for d in donor_db.keys()}
+
+def view_donor_names():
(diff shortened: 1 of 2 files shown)


students/alex_whitty/session06/mailroom4.py
```diff
@@ -0,0 +1,171 @@
...
+    print()
+    print("{:20} | {:10} | {:10} | {}".format(headers[0], headers[1], headers[2], headers[3]))
+    print("-" * 80)
+
+    for k, v in donor_db.items():
+        total = (sum(v))
+        times = (len(v))
+        avg = (sum(v) / len(v))
+        summary.append([k, total, times, avg])
+    summary.sort(key=lambda d: d[1], reverse=True)
```
mdgregor: You could break out this for loop to a function for unit testing.

students/alex_whitty/session06/mailroom4.py
```diff
@@ -0,0 +1,171 @@
...
+            return value
+        except ValueError:
+            print("Invalid value. Please try again")
+            continue
+
+def current_donor(donor):
+
+    amount = get_value("Please enter donor amount.>>>", float)
+    donor_db[donor].append(float(amount))
+    dir_for_letter()
```
mdgregor: you could probably mock this function

students/alex_whitty/session06/mailroom4.py
```diff
@@ -0,0 +1,171 @@
...
+    with open(f'{donor}.txt', 'wt') as letter:
+        letter.write(switch_func_dict.get(0)(donor, amount))
```
mdgregor: I'd then open this file and make sure the donor go written to the file.

mdgregor:
The code should work exactly the same as it did before refactor (i.e. the user shouldn't know the difference).  If you want to test the printing of reports and letters you can write system's standard out to a file and then take it as input for the test.
```
>>> from contextlib import redirect_stdout
>>> with open('test_out.txt', 'w') as f:
...     with redirect_stdout(f):
...             print("here is some stuff")
... 
>>> exit()
ferros@desktop-ferros:~$ ls
CLionProjects  Desktop  Documents  Downloads  evacsouth.pdf  Music  Pictures  Programs  Public  PycharmProjects  retropie  Templates  test_out.txt  Videos  vpn-config
ferros@desktop-ferros:~$ cat test_out.txt 
here is some stuff
```
Source: https://stackoverflow.com/questions/4675728/redirect-stdout-to-a-file-in-python

PR #136, opened 2019-02-25
Lesson06 scripts

mmwheel:
Lesson06 scripts are ready for review

### DIFF
--- a/students/marsha_w/lesson_06/test_mailroom.py
+++ b/students/marsha_w/lesson_06/test_mailroom.py
@@ -1,0 +1,100 @@
+#!/usr/bin/env python3
+
+'''Title: Unit testing for create a report function
+Dev: Marsha Wheeler
+Date: 02/23/2019
+'''
+
+from mailroom4 import create_report, write_text, write_file, send_thank_you, donor_dict
+import os.path
+import datetime
+
+# test create report function
+def test_donationsum_order():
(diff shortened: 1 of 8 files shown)


students/marsha_w/lesson_06/test_mailroom.py
```diff
@@ -0,0 +1,100 @@
...
+    for keys in donor_dict:
+        date = str(datetime.datetime.now()).split()
+        out_name = str(keys.replace(' ', '_') + '_' + date[0] + ".txt")
+        assert os.path.isfile(
+            '/Users/marshawheeler/GP_Python210B_Winter_2019/students/marsha_w/lesson_06/' + out_name)
+
+
+#test appending send_thank_you function
+
+def test_donor_name1(donor="Paul Allen", amount=float(500)):
```
mdgregor: not sure why you included donor, amount as function variables here.  What would have been interesting is if you abstracted this idea out for testing.  For example:
```
def _validate_thank_you(donor, amount):
    send_thank_you(donor, amount)
    assert donor in donor_dict
```
And have your tests call that function passing in the donor and mount.  Keep in mind though, you only need to test it once rather than with multiple donors and amounts each time, as you don't really gain anything by testing the same functionality 4-5 times if that makes sense from a programming perspective, unless those inputs where slightly different each time.

PR #138, opened 2019-02-25
Submitting Session06

awhitty1:
Submitting session 6 files.

### DIFF
--- a/students/alex_whitty/session06/mailroom4.py
+++ b/students/alex_whitty/session06/mailroom4.py
@@ -11,7 +11,6 @@ donor_db = {"William Gates, III": [653772.32, 12.17],
             }
 
 
-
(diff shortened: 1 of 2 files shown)


mdgregor:
Hey Alex.  So this test code doesn't actually excercise the key parts of your mailroom program.  For each of them your doing `self.assert` on variables that your creating in the test and not part of mailroom.  For the letters your also just asserting that they return something via the `True` rather than asserting that the string is properly created when you pass in the donor and amount.

For example.  I could remove everything in the `write_a_letter` functions and just say `return 1` and your unit test would pass.  Our goal with the unit tests is to verify that despite the changes we make to the program, that it still operates as intended.  Have a read through this resource and hit me up with specific questions as you have them. https://pythontesting.net/start-here/

PR #140, opened 2019-02-25
added mailroom3

bbirsoz1:


students/bbirsoz1/Module_5/mailroom3.py
```diff
@@ -0,0 +1,126 @@
...
+    current_d = "" #donors currently listed
+    current_d = str(input("Who would you like to mail?:"))
+    if current_d in donor_db:
+        add_donation(current_d)
+    else:
+        donor_db[current_d] = []
+        add_donation(current_d)
+
+def add_donation(current_d):
+    donation_times = float(input("How many donations have been made so far?: ")) #replaced int() with float()
```
ldconejo: This one should not be a _float_, as the number of donations will always be an integer. This is also a good location to add exception handling, in case the user decides to enter a non-numeric value, such as "none" that will make your code crash.

students/bbirsoz1/Module_5/mailroom3.py
```diff
@@ -0,0 +1,126 @@
...
+    while True:
+        try:
+            donation_times = int(donation_times)
+            break
+        except ValueError:
+            print("Input must be an integer, try again.") #it doesnt show up, why???
+            break
+    total_donation = 0
+    while donation_times > 0:
+        new_donation = float(input("Donation amount?: "))
```
ldconejo: This is also a place where you should add exception handling, since your code will crash if a non-numeric string is entered.

students/bbirsoz1/Module_5/mailroom3.py
```diff
@@ -0,0 +1,126 @@
...
+        elif mail == "1":
+            mail_donor()
+        elif mail == "2":
+            list_of_donors()
+        elif mail == "3":
+            break
+
+def main():
+    while True:
+        response = int(input(main_menu_prompt))
```
ldconejo: Add exception handling here as well. It warns you if you enter a number that doesn't match any menu items, but entering a string crashes your code.

ldconejo:
Hi Begum,

Another thing: Try adding docstrings to your functions to indicate what they do. For example:

```
def list_of_donors():
    '''
    Shows the current list of donors
    '''
```

This will improve readability of your code.

Regards,

Luis

PR #141, opened 2019-02-26
Module 6 files

sdlipp:
Mailroom_Part4

I'm not sure if I set up all the tests as intended.  Let me know if I missed the mark a bit so I can make some adjustments.

### DIFF
--- a/students/ScottL/session06/Mailroom_Part4.py
+++ b/students/ScottL/session06/Mailroom_Part4.py
@@ -1,0 +1,217 @@
+#!/usr/bin/env python3
+
+import sys
+
+global_donation_list = {"Tom Petty": [5000, 15000],
+                        "Neil Diamond": [20000, 30000],
+                        "Elton John": [25000, 25000, 40000],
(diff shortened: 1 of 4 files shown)


ldconejo:
Hi Scott,

Your tests look good. The assignment requires _full suite_ which in the ideal case would mean closer to 100% coverage. A more realistic number would be 90%.

You can test coverage by using the _coverage_ module with the following command:

```
$ python -m coverage run --include=Mailroom_Part4.py -m pytest test_Mailroom_Part4.py
```

After that, check out the coverage report with this command:

```
$ python -m coverage report
```

In your case, the report shows the following:

```
Name                Stmts   Miss  Cover
---------------------------------------
Mailroom_Part4.py      99     43    57%
```

See if you can add some additional tests or modify the existing ones to "touch" additional segments of your code that are currently not covered.

Finally, give Pylint a run on both your mailroom code and your tests and fix some PEP8 issues there.

Regards,

Luis

PR #142, opened 2019-02-26
Submitting Mailroom4

mjchang14:
That was a lot harder than I thought it would be! I was very tempted to work "backwards" in creating the tests, but I can see how it makes it easier to write the tests first, then write the program.


ldconejo:
Hi Marisa,

One of the requirements of the assignment is to test the full suite, meaning to get as close as possible to running 100% of your code. A good approximation would be 90%. One way to check that is using the _coverage_ module which you can run, combined with _pytest_ with the following command:

```
$ python -m coverage run --include=mailroom4.py -m pytest
```

Once you run this, you can check the coverage report with this command:

```
$ python -m coverage report
```

In you case, the coverage report shows this:

```
Name           Stmts   Miss  Cover
----------------------------------
mailroom4.py      82     59    28%
```

Several of your tests, for example, focus on the _donor_db()_ function, leaving other functions in your code with no coverage. 

_def test_gen_letters()_, on the other, doesn't test any code at all, just the existence of files that have to be created manually before any testing is done. This test should be modified to execute _thank_you_all()_ and then check the existence of the files that function was expected to generate.

Try to implement some additional tests and/or modify the existing ones, while check progress with coverage percentage as indicated above.

Regards,

Luis

ldconejo:
Another recommendation is to run Pylint on both _mailroom4.py_ and _test_mailroom4.py_ and correct a few issues reported (including unused imports).

mjchang14:
Thank you for all your help and patience, Luis! I think in my effort to get
caught up I've been going in circles in my head, missing things, and not
processing everything fully. I'll check out the coverage module and run
Pylint, then work on fixing this and will submit more functional code.

Thanks!

-Marisa

PR #144, opened 2019-02-28
html_render working with 100% test coverage.  Looking for feedback / things to improve

Douglas-Klos:
This was a tricky assignment but I think I've got my head wrapped around it.

The test file should be covering 100% of the code in html_render.py, at least according to pytest-cov.  All classes defined in html_render are being tested.

Are there areas that I can improve upon?

pylint seems to complain about:
__init__ method from base class 'Element' is not called (super-init-not-called)
There are a couple instances where I do call super init, but other times it's not wanted as I'm trying to avoid part of the code in super init.  Is pylint being over zealous, or should you always be calling super init?

The other pylint message I'm getting a lot of is:
Method 'append' is abstract in class 'SelfClosingTag' but is not overridden (abstract-method)
Also for OneLineTag.  Is this relevant?  Why would I need to override append for these?

:edit: Flake8 doesn't seem to care about either of the above.

Thanks for the insight, I appreciate the feedback.


students/douglas_klos/session7/html_render.py
```diff
@@ -1,18 +1,163 @@
...
-        pass
+class Element():
+    """ Framework for a basic element in HTML code """
+    tag = 'html'
+    indent = ''
+
+    def __init__(self, content=None, **kwargs):
+        self.attributes = kwargs
+
+        if content is not None:
```
ldconejo: _None_ always returns False when you test for truth, so you can simplify this line:

```
if not content:
```

A few more information on which values test False in Python (an older document, but still holds):

https://docs.python.org/2.4/lib/truth.html

ldconejo:
Hi Doug,

The output looks good and coverage is indeed at 100%. Very impressive.

I added a comment on testing _None_ for truth, you should change that in your code.

On the Pylint warnings, Python does not require you to call the init method of the super class every time. Same goes for not overriding the append method.

For warnings that you know are not a real issue within your code, you can tell Pylint to override them either for an entire module or individual classes. In this case, you can add this to line 2 of _htm_render.py_:

```
#pylint: disable=W0223,W0231
```

This can be a little counter-intuitive, as you would normally think of comments as being ignored by default. In this case, Pylint will parse that comment and disable those two warnings.

Pylint is also flagging the lack of docstrings for some of your class methods. Try to fix those.

Regards,

Luis

PR #145, opened 2019-03-01
Stuck

UncleanlyCleric:
I cannot figure out how to complete the 12th test:

=================================== FAILURES ===================================
_______________________________ test_attributes ________________________________

    def test_attributes():
        '''
        Testing attribute passing
        '''
        e = P('A paragraph of text', style='text-align: center', id='intro')
    
        file_contents = render_result(e).strip()
        print(file_contents)  # so we can see it if the test fails
    
        # note: The previous tests should make sure that the tags are getting
        #       properly rendered, so we don't need to test that here.
        #       so using only a 'P' tag is fine
        assert 'A paragraph of text' in file_contents
        # but make sure the embedded element's tags get rendered!
        # first test the end tag is there -- same as always:
        assert file_contents.endswith('</p>')
    
        # but now the opening tag is far more complex
        # but it starts the same:
        assert file_contents.startswith('<p')
>       assert 'style="text-align: center"' in file_contents
E       assert 'style="text-align: center"' in '<p>\nA paragraph of text\n</p>'

test_html_render.py:250: AssertionError

I think part of this is because I really can't wrap my head around what I'm doing here.  Is there any chance someone could look and tell me where I'm going wrong?


ldconejo:
Hi Jesse,

Note the test is passing attributes to the init method:

```
def test_attributes():
    '''
    Testing attribute passing
    '''
    e = P('A paragraph of text', style='text-align: center', id='intro')
```

The _init_ method in the _Element_ class will be getting those as kwargs:

```
    def __init__(self, content=None, **kwargs):
```

_'A paragraph of text'_ will be assigned to _content_. You need to create an instance variable to store the values being passed with kwargs.

Now that you have those values available to your instance, you need to figure out when and where to add them. Consider using the _render_ method as the place where you would incorporate those attributes into the tag.

Hope this helps!

-Luis

PR #152, opened 2019-03-04
Session07 - Unfinished

pauleclifton:
I was not able to completed the assignment beyond part 4. Following the tutorial, I could not figure out what I needed to do to the render functions to make them work.


students/paul_c/session07/html_render.py
```diff
@@ -0,0 +1,64 @@
...
+        if content is not None:
+            self.contents = [content]
+
+    def append(self, new_content):
+        self.contents.append(new_content)
+
+    def render(self, out_file):
+        open_tag = ["<{}".format(self.tag)]
+        for key, value in self.attributes.items():
+            open_tag.append(' ' + key + '="' + str(value) + '"')
```
ldconejo: This works, but you should add the closing ">" to the tag, after you have added all attributes (if any):
```
        open_tag.append(">\n")
```

students/paul_c/session07/html_render.py
```diff
@@ -0,0 +1,64 @@
...
+
+class OneLineTag(Element):
+
+    def render(self, out_file):
+        open_tag = ["<{}".format(self.tag)]
+        for key, value in self.attributes.items():
+            open_tag.append(' ' + key + '="' + str(value) + '"')
+        out_file.write("".join(open_tag))
+
+        out_file.write("<{}>".format(self.tag))
```
ldconejo: This will cause the tag to be added again, for example, for title, it will end up looking like this:

`<title<title>PythonClass = Revision 1087:</title>`

students/paul_c/session07/html_render.py
```diff
@@ -0,0 +1,64 @@
...
+    tag = 'head'
+
+class OneLineTag(Element):
+
+    def render(self, out_file):
+        open_tag = ["<{}".format(self.tag)]
+        for key, value in self.attributes.items():
+            open_tag.append(' ' + key + '="' + str(value) + '"')
+        out_file.write("".join(open_tag))
+
```
ldconejo: Again in this case, you need to close the tag, which could be achieved by adding:

`        open_tag.append(">")`

Note that there is no carriage return, as this is a one-liner.

ldconejo:
Hi Paul,

Your code is actually working rather well. I made a few comments on changes you can make to get it all working, up to and including step #4. At that point, you will need to add some more code to get to step #5 and further, but you have already done a lot of the heavy lifting in your code.

Another thing pending is indentation, you can give it a try or live it for the end (if you can do it earlier, it will make your HTML output easier to read and debug).

Let us know if you have any questions!

-Luis

PR #161, opened 2019-03-06
session 08 Circle Class Question

yiyunxu:
1. For step 5, how to create an alternative constructor using classmethod?
2. For step 7, c2 * 3 works. but what about 3 * c2 ? – how to fix that?

### DIFF
--- a/students/elaine_x/session08/circle_exercise_ex.py
+++ b/students/elaine_x/session08/circle_exercise_ex.py
@@ -1,0 +1,86 @@
+'''
+##########################
+#Python 210
+#Session 07 - Circle Exercise
+#Elaine Xu
+#Mar 5, 2019
+###########################
+'''
+import math
+#############################
+#Step 1 create class called circle
+#############################
+class Circle(object):
+    def __init__(self, radius):
+        self.radius = radius
+
+#############################
+#Step 2 add diameter property
+#############################
+    @property
+    def diameter(self):
+        print("in getter")
+        return self.radius * 2
+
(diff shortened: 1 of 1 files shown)


students/elaine_x/session08/circle_exercise_ex.py
```diff
@@ -0,0 +1,86 @@
...
+#############################
+# Step 5 add an alternate constructor that create circle directly with the diameter
+#############################
+
+#    @classmethod
+#    def from_diameter(cls, diameter):
+#        self = cls(self)
+#        self.diameter = diameter
+#        self.radius = diameter/2
+#        return self.diameter
```
mdgregor:
```
@classmethod
def from_diameter(cls, diameter):
    radius = diameter / 2
    return cls(radius)
```

students/elaine_x/session08/circle_exercise_ex.py
```diff
@@ -0,0 +1,86 @@
...
+#############################
+'''c = Circle(4)
+c = Circle.from_diameter(8)
+print(c.diameter)
+'''
+c1 = Circle(2)
+c2 = Circle(4)
+print(c1+c2)
+print(c2 * 3)
+#print(3 * c2)
```
mdgregor:
```
    def __mul__(self, value):
        try:
            return self.radius * value
        except TypeError:
            __rmul__(self, value)

    def __rmul__(self, value):
        return self.radius * value
```
The key is to catch the type error.

PR #169, opened 2019-03-08
Code for Review

Mahmood08:
Could you please check what I am missing here? My three test adding , multiplying and sorting circles keeps failing.
Thank you

### DIFF
--- a/students/shirin_ak/Session08/Circle.py
+++ b/students/shirin_ak/Session08/Circle.py
@@ -1,9 +1,11 @@
 
 import math
+
 class Circle():
     def __init__(self, radius):
         # STEP 1 The radius is a required parameter
         self.radius = radius
+        
     # STEP 2 Add a 'diameter' property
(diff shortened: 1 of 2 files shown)


mdgregor:
> Could you please check what I am missing here? My three test adding , multiplying and sorting circles keeps failing.
> Thank you

It because you are testing to see if the objects are the same, rather than a value.  for example:
```
>>> from circle import Circle
>>> c1 = Circle(6)
>>> c2 = Circle(6)
>>> c1 == c2
False
>>> id(c1)
140043251196256
>>> id(c2)
140043251196480
>>> 
```
You can see above, even though we instantiated two circles with the same value,  they are separate objects, so will never be equal

You'll want to update your tests to test for a value returned.  i.e. the following:
```
>>> (c1 + c2).radius == 12
True
```

Mahmood08:
Thank you. It works.

PR #170, opened 2019-03-10
Assignment 8 files

sdlipp:
Circle class and tests

### DIFF
--- a/students/ScottL/session08/circle_class.py
+++ b/students/ScottL/session08/circle_class.py
@@ -1,0 +1,84 @@
+
+"""
+Homework 8 - Circle Class
+Use objects and their associated methods for characterizing circles and spheres.
+"""
+
+import math
+
+
+class Circle:
+    """Define and quantify properties for circles with a given radius"""
+
+    def __init__(self, radius):
+        self.radius = radius
+
+    @property
+    def diameter(self):
+        """Return the diameter calculated from the radius property."""
+        return self.radius * 2
+
+    @diameter.setter
+    def diameter(self, value):
+        self.radius = value / 2
+
+    @property
+    def area(self):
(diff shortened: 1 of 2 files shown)


students/ScottL/session08/circle_class.py
```diff
@@ -0,0 +1,84 @@
...
+        return cls(diameter / 2)
+
+    def __str__(self):
+        return f"Circle with a radius of: {self.radius}"
+
+    def __repr__(self):
+        return f"Circle({self.radius})"
+
+    def __add__(self, other):
+        return f"Circle({self.radius + other.radius})"
```
mdgregor: This should result in a `Circle()` object, not a string value

students/ScottL/session08/circle_class.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def __mul__(self, other):
+        return f"Circle({self.radius * other})"
```
mdgregor: same here.

students/ScottL/session08/circle_class.py
```diff
@@ -0,0 +1,84 @@
...
+        return f"Sphere with a radius of: {self.radius}"
+
+    def __repr__(self):
+        return f"Sphere({self.radius})"
+
+    def __add__(self, other):
+        return f"Sphere({self.radius + other.radius})"
+
+    def __mul__(self, other):
+        return f"Sphere({self.radius * other})"
```
mdgregor: these two as well

students/ScottL/session08/circle_class.py
sdlipp: ahh yeah that makes more sense, i'll get that adjusted

PR #171, opened 2019-03-10
Submitting Session08-Circle class

Mahmood08:


students/shirin_ak/Session08/Circle.py
```diff
@@ -35,16 +44,20 @@ def __str__(self):
...
    #STEP 7 Add two circles and multiply one by a number
     def __add__(self, other):
       
-        return Circle(self.radius + other.radius)
+        return self.radius + other.radius
     
     def __mul__(self, other):
-        return Circle(self.radius * other)
-    
+        return self.radius * other
```
mdgregor: so you had it right the first type returning the object, but for your test you want to test against the value of the object.  
For example:
```
def __mul__(self, other):
    return self.__class__(self.radius + other.radius)
```

Then for the test:
```
def test_mult():
    c1 = Circle(5)
    actual = c1 * 3
    expected = 15
    assert actual.radius == expected
```

PR #172, opened 2019-03-10
session08 hw submit

yiyunxu:


students/elaine_x/session08/circle_exercise_ex.py
```diff
@@ -58,29 +54,52 @@ def __str__(self):
     def __repr__(self):
         return "Circle({})".format(self.radius)
 
-
 #############################
 # Step 7 add numerical protocol
 #############################
     def __add__(self, other):
-        return "Circle({})".format(self.radius + other.radius)
+        return (self.radius + other.radius)
```
mdgregor: so you had it right the first type returning the object, but for your test you want to test against the value of the object.  
For example:
```
def __mul__(self, other):
    return self.__class__(self.radius + other.radius)
```

Then for the test:
```
def test_mult():
    c1 = Circle(5)
    actual = c1 * 3
    expected = 15
    assert actual.radius == expected
```

PR #174, opened 2019-03-11
Session 08

dac2229:
Finished week 08 assignments.

### DIFF
--- a/students/daniel_carrasco/session08/circle.py
+++ b/students/daniel_carrasco/session08/circle.py
@@ -1,10 +1,84 @@
-class Circle():
-
+import math
+class Circle:
+    pi = math.pi
     def __init__(self,the_radius):
         self.radius = the_radius
-        self.diameter = self.radius*2
+
+    @property
+    def diameter(self):
+        return self.radius*2
+
+    @diameter.setter
+    def diameter(self, diameter):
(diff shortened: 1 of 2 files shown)


students/daniel_carrasco/session08/circle.py
```diff
@@ -0,0 +1,105 @@
+import math
+
+
+class Circle(object):
+    pi = math.pi
```
mdgregor: you can do `from math import pi` and you don't need this line.

students/daniel_carrasco/session08/test_circle.py
```diff
@@ -0,0 +1,96 @@
...
+def test_str_repr():
+    "test to see if __str__ and __repr__ methods work"
+    a = str(Circle(4))
+    b = repr(Circle(5))
+    c = str(Sphere(3))
+    d = repr(Sphere(2))
+    assert a == 'Circle with a radius of 4'
+    assert b == 'Circle(5)'
+    assert c == 'Sphere with a radius of 3'
+    assert d == 'Sphere(2)'
```
mdgregor: You don't need these extra variables.  you can just do:
```
assert str(Circle(4) == 'Circle with a radius of 4'
```

PR #175, opened 2019-03-11
Submitting Session08 files

awhitty1:
Submitting session08 files(circle and test_circle).

### DIFF
--- a/students/alex_whitty/session08/test_circle.py
+++ b/students/alex_whitty/session08/test_circle.py
@@ -1,0 +1,80 @@
+from math import pi
+from circle import Circle, Sphere
+
+
+def test_radius():
+    c = Circle(2)
+    assert c.radius == 2
+
+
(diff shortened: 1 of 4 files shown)


awhitty1:
Hello Michael,

Could you please review my session 06 files, mailroom4 & test_mailroom. Updated the test results.

students/alex_whitty/session08/test_circle.py
```diff
@@ -0,0 +1,85 @@
...
+
+def test_string_repr():
+    c = Circle(12)
+    assert repr(c) == "Circle(12)"
+
+
+def test_less():
+    c1 = Circle(2)
+    c2 = Circle(5)
+    assert (c1 < c2) is True
```
mdgregor: when using `assert` you'll want to just use the operators, no need for the `is True`

PR #177, opened 2019-03-11
Lesson 8 sumbisson

woosukjeung:


students/WooseokJ/Lesson 8/CircleWJ.py
```diff
@@ -0,0 +1,151 @@
...
+        return cls(value/2)
+
+    def __str__(self):
+        return "circle with radius: {:.4f}".format(self.radius)
+
+    def __repr__(self):
+        return "circle({})".format(self.radius)
+
+    def __add__(self, other):
+        return self.radius + other.radius
```
ldconejo: This should return a Circle object that has the combined radius of the two circles added, not just the sum of the two radius.

ldconejo: The same applies to your implementation of _rmul_, _mul_ and _div_. They should all return new Circle objects with the appropriate radius.

students/WooseokJ/Lesson 8/CircleWJ.py
```diff
@@ -0,0 +1,151 @@
...
+    c1 = Circle(2)
+    c2 = Circle(4)
+    print(c1 + c2)
+    assert (c1 + c2) == Circle(6)
+
+
+def test_mul():
+    c1 = Circle(4)
+    assert (c1 * 3) == Circle(12)
+
```
ldconejo: Note that _test_add()_ and _test_mul()_ are both failing due to your _add_ and _mul_ methods returning integers instead of _Circle_ objects, so there's not _radius_ attribute to check the assertion:

```
===================================================================== FAILURES ======================================================================
_____________________________________________________________________ test_add ______________________________________________________________________

    def test_add():
        c1 = Circle(2)
        c2 = Circle(4)
        print(c1 + c2)
>       assert (c1 + c2) == Circle(6)

CircleWJ.py:128: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = circle(6), other = 6

    def __eq__(self, other):
>       if self.radius == other.radius:
E       AttributeError: 'int' object has no attribute 'radius'

CircleWJ.py:54: AttributeError
--------------------------------------------------------------- Captured stdout call ----------------------------------------------------------------
6
_____________________________________________________________________ test_mul ______________________________________________________________________

    def test_mul():
        c1 = Circle(4)
>       assert (c1 * 3) == Circle(12)

CircleWJ.py:133: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = circle(12), other = 12

    def __eq__(self, other):
>       if self.radius == other.radius:
E       AttributeError: 'int' object has no attribute 'radius'

CircleWJ.py:54: AttributeError
```

students/WooseokJ/Lesson 8/CircleWJ.py
```diff
@@ -0,0 +1,151 @@
...
+        return self.radius * other
+
+    def __div__(self, value):
+        return self.radius / value
+
+# allow to compre between two circles
+    def __lt__(self, other):
+        if self.radius < other.radius:
+            return True
+        else:
```
ldconejo: This _else_ statement is unnecessary, since it follows a return. The code for this method could be simplified to:

```
    def __lt__(self, other):
        if self.radius < other.radius:
            return True
        return False
```

students/WooseokJ/Lesson 8/CircleWJ.py
```diff
@@ -0,0 +1,151 @@
...
+            return False
+
+    def __eq__(self, other):
+        if self.radius == other.radius:
+            return True
+        else:
```
ldconejo: This _else_ statement is also unnecessary.

ldconejo:
Hi Wooseok,

A couple of tests are failing and you need to modify your implementation of arithmetic operations with Circle objects to return Circle objects instead of integers. Please submit a new pull request once those are ready. 

Regards,

Luis

PR #178, opened 2019-03-11
committing lesson08 for review

mmwheel:


students/marsha_w/lesson_08/circle.py
```diff
@@ -0,0 +1,56 @@
...
+    @property
+    def area(self):
+        area = (self.diameter / 2)**2 * math.pi
+        return area
+
+    @classmethod
+    def from_diameter(cls, diameter):
+        cls.diameter = diameter
+        cls.radius = diameter / 2.0
+        return cls
```
mdgregor: you just want to do:
```
@classmethod
    def from_diameter(cls, diameter):
        return cls(diameter / 2)
```

students/marsha_w/lesson_08/circle.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def __add__(self, other):
+        sumC = self.radius + other.radius
```
mdgregor: no camel case variable names.

students/marsha_w/lesson_08/circle.py
```diff
@@ -0,0 +1,56 @@
...
+        return sumC
+
+    def __mul__(self, val):
+        mulC = self.radius * val
+        return mulC
+
+    def __rmul__(self, other):
+        mulC = self.radius * other
+        return mulC
```
mdgregor: you actually don't need these variables anyway, just to `return self.radius * other`

PR #179, opened 2019-03-12
submit Sparse Array for code review

yiyunxu:


students/elaine_x/session08/sparse_array_ex.py
```diff
@@ -0,0 +1,127 @@
...
+    '''generates sparse array'''
+    def __init__(self, array):
+        self.array = {}
+        self.len = len(array)
+        for i, item in enumerate(array):
+            if item is not 0:
+                self.array[i] = item
+
+    def __str__(self):
+        return "Sparse Array({})".format(self.array)
```
ldconejo: This one should have a return that looks more like a list, as the sparse array acts as compression for lists with lots of zero values. You could also consider implementing the _repr_ method and have it return a list.

students/elaine_x/session08/sparse_array_ex.py
```diff
@@ -0,0 +1,127 @@
...
+Can you make it support slicing?
+How else can you make it like a list?
+'''
+class SparseArray():
+    '''generates sparse array'''
+    def __init__(self, array):
+        self.array = {}
+        self.len = len(array)
+        for i, item in enumerate(array):
+            if item is not 0:
```
ldconejo: Use _if item != 0_ instead.

students/elaine_x/session08/sparse_array_ex.py
```diff
@@ -0,0 +1,127 @@
...
+        if isinstance(i, slice):
+            print(i)
+            if i.stop > self.len:
+                raise IndexError
+            self.slicearray = {}
+            for ind in range(i.start, i.stop):
+                if ind in self.array:
+                    self.slicearray[ind] = self.array[ind]
+            return self.slicearray
+        else: #simple index integer
```
ldconejo: This _else_ is unnecessary, as it is preceded by a _return_ statement.

students/elaine_x/session08/sparse_array_ex.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def __str__(self):
+        return "Sparse Array({})".format(self.array)
+
+    def __getitem__(self, i):
+        if isinstance(i, slice):
+            print(i)
+            if i.stop > self.len:
+                raise IndexError
+            self.slicearray = {}
```
ldconejo: This is fine, but whenever possible, try to initialize all your instance attributes, such as _self.slicearray_ in this case, in your _init_ method. The issue here is that calling _slicearray_ before _getitem_ has been called at least once will cause your program to error out, since the attribute has not been initialized yet.

ldconejo:
Hi Elaine,

I added some comments on things that could be improved. Code looks good in general. Since this is not a graded assignment, I'll leave up to you to submit a new pull request with improvements.

Regards,

Luis

PR #181, opened 2019-03-13
Mailroom OO working, submitting

Douglas-Klos:
I believe this covers all the requirements (and more) for the object oriented mailroom program.  I wanted to make sure that this code has been structured and factored properly (appropriate use of classes and data structures, organization, etc.)  Please let me know if there's anything that should be changed / altered / implemented differently.  

Execution examples:
Basic program: ./mailroom.py
Show CLI help menu: ./mailroom.py -h
Load database file mailroom_db.pkl: ./mailroom.py --open mailroom_db.pkl

Linting line width is set to 100.

Thanks for the input!


students/douglas_klos/session9/mailroom/donor.py
```diff
@@ -0,0 +1,117 @@
...
+class Donor():
+    """
+    Donor class
+
+    Attributes:
+        name : The name of the donor
+        donations : The donations for the donor
+    """
+
+    THANK_YOU_LETTER = ('Dear {}:\n'
```
ldconejo: This constant probably belongs as part of the user interface. Only Donor information should be here.

ldconejo: Of course, the argument could be made that you want the same thank you letter displayed, regardless of what the user interface is.

students/douglas_klos/session9/mailroom/donordb.py
```diff
@@ -0,0 +1,247 @@
...
+                report += f'\t\t0    '
+            else:
+                report += f'{len(self.database[key].donations):10}    '
+            report += f'$ {self.database[key].average_donation:14,.2f}\n'
+
+        report += "-" * 79 + '\n'
+
+        return report
+
+    def html_report(self):
```
ldconejo: Fancy :-) 
This is good, still, any data visualization code should be separated from the data itself.

ldconejo:
Hi Doug,

Code looks very good, the work done on data visualization goes well beyond the assignment requirements, being capable to display the data in HTML format.

One thing to consider is separating any data visualization and data storage code from you data structure. This will make your code easier to maintain when you have to support multiple platforms.

-Luis

Douglas-Klos:
Thank you for the feedback.  I've had a lot of fun in this class, and with this assignment.  I went back and forth, to have the thank you letters in the donordb vs the UI - I placed it in the donordb for ease of coding but I can see why you'd want it to be in the users territory so that they could be changed easier.  I'll work on moving those over.

With the file writing, it seemed like 'logic' code, that's why I placed it in donordb.  _In a production environment, would it be more typical to move file operations like this to the UI code, or should the donordb class have OS detection and write the file appropriately for windows/nix?_

I'll work on making these changes, I'll try and get it corrected by the deadline tonight.  Thanks again!

ldconejo:
Excellent work, Doug. Test coverage for _donor.py_ and _donordb.py_ is at 100%, you even have 78% coverage for _mailroom.py_ (which wasn;'t a requirement, but it's still very good) at Pylint grade of 10 for all three files.

-Luis

PR #182, opened 2019-03-14
Pull request

UncleanlyCleric:
Submitting this now.  I'm no where near done, I just want to get this started in case I need help.


students/jesse_miller/session09/cli_main.py
```diff
@@ -0,0 +1,171 @@
...
+    def donor_mail_choice():
+        '''
+        This section allows the user to mail a donor
+        '''
+        current_donor = ''
+        MailRoom.list_donors()
+        try:
+            donor = str(input('Who would you like to mail (all for all): '))
+            if alms.find_donor(donor):
+                MailRoom.mail_send_one(donor)
```
ldconejo: Note that _donor_ here is the string value you're getting from the your _input_ call on line 62. You're then calling _mail_send_one()_, which in turn calls _Donor.letter_template(donor)_. 

Now, _letter_template(self)_ is supposed to be an instance method but the way you're calling it, it's behaving like a function that takes one parameter, _self_, which you're loading with _donor_, a string value (that's the reason for the "str object has no attribute name'). The way _letter_template_ is configured, it should be a Donor object, which would have a _name_ attribute.

ldconejo: Another consideration to make is that _letter_template_ has no place in _donor_models.py_, because it deals with data visualization, not with data structure. _letter_template_ should instead be in _cli_main.py_.

One alternative would be to get _donor_dict_ from _cli_main.py_ and iterate through it, printing the thank you letter for each donor, sum(donations).

Another one would be to have a new instance method in _donor_models.py_ that does something like this:

def get_donor_total_donations(donor_name):
    return sum(self.donors_dict[current_donor])

Then you can call it from _cli_main.py_ when printing out thank you letters from each donor.

students/jesse_miller/session09/donor_models.py
```diff
@@ -0,0 +1,154 @@
...
+    Find a donor
+    List donors
+    Donor reports
+    Data saving and reloading.
+    '''
+    def __init__(self):
+        '''
+        Initializing and building the default dict
+        '''
+        self.donors_dict = {'Robert Smith': [435.56, 125.23, 357.10],
```
ldconejo: _self.donors_dict_ should instead have the key (the name of the donor is fine) point to a _Donor_ object, instead of pointing to a list of donations. This will make it easier to add new donor _features_ by simply modifying the _Donor_ class.

UncleanlyCleric: So:
donor = "firstname lastname"

donor.donors_dict (functions)

Am I understanding this right?

UncleanlyCleric: Well, 

total = Donors(donor.donations_total) 

for example?

students/jesse_miller/session09/donor_models.py
```diff
@@ -0,0 +1,154 @@
...
+                            'Devin Townsand': [431.12, 342.92, 5412.45],
+                           }
+
+
+    def donor_creation(self, current_donor):
+        '''
+        Allows for adding new donors to the db
+        '''
+        new_donor = Donor(current_donor)
+        self.donors_dict[current_donor] = new_donor
```
ldconejo: This is good, since it handles _self.donors_dict_ as a dictionary containing _Donor_ objects.

students/jesse_miller/session09/donor_models.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def find_donor(self, current_donor):
+        '''
+        This will take an inputed donor name, and search for it.  If the name
+        isn't found, it will then bounce the request to the creation function.
+        '''
+        try:
+            return self.donors_dict[current_donor]
+        except KeyError:
+            return self.donor_creation(current_donor)
```
ldconejo: Do it the other way around, have _donor_creation_ call _find_donor_ to see if a donor already exists before trying to create a new one.

students/jesse_miller/session09/donor_models.py
```diff
@@ -0,0 +1,154 @@
...
+class Donor:
+    '''
+    This will be slightly easier in a sense because we're only dealing with a
+    single donor.  However, I get to look forward to pairing everything down to
+    an interesting, if not insane, degree from previous attempts.
+    '''
+
+    def __init__(self, name):
+        self.name = name
+        self.donations = [] #?  We'll try it.
```
ldconejo: This is ok. In a more complex scenario, this one could become a list of _Donation_ objects, with more details about each donation.

students/jesse_miller/session09/donor_models.py
ldconejo: Not exactly.

Let's simplify the initialization, just do:

```
self.donors_dict = {}
```
Then, when you add a new donor, the process will go like this:

1. Ask the user for the name & last name of the new donor.
2. Check if that donor already exists in _donors_dict_.
3. Call _donor_creation_, which will create a new _Donor_ object and add it to _donors_dict_ inside of the _DonorCollection_ object.

ldconejo: Right now, the data in the dictionary is not consistent, since the example you're loading has each key pointing to a list of integers, not to a Donor object.

students/jesse_miller/session09/cli_main.py
```diff
@@ -0,0 +1,180 @@
...
+from donor_models import DonorCollection, DonorTools
+from mail_box import MailBox
+'''
+Module imports
+'''
+
+#pylint: disable=C0103
+name = ''
+alms = DonorCollection()
+tools = DonorTools(name)
```
ldconejo: No need for this one.

UncleanlyCleric:
Okay, I definitely need another set of eyes.  As of this point, everything is working except adding donations to a new donor.  Right now, it's writing the memory register and not the donation total.

students/jesse_miller/session09/cli_main.py
UncleanlyCleric: Yeah, that's a legacy from me experimenting with different ideas.  :\

students/jesse_miller/session09/cli_main.py
```diff
@@ -0,0 +1,180 @@
...
+        try:
+            MailRoom.list_donors()
+            donor = str(input('Enter the name of the donor to add: '))
+            alms.donor_creation(donor)
+            while True:
+                d_num = int(input('How many donations were made: '))
+                while d_num > 0:
+                    donations = float(input('Enter their donation: '))
+                    #self.donors_dict[donor].append(new_don)
+                    tools.donation_add(donations)
```
ldconejo: You need to call _donations_add_ from the new Donor that you have created, this line should be:

```
alms.donors_dict[donor].donation_add(donations)
```

UncleanlyCleric: I had tried that once, this is what I get:

'Kerry': <donor_models.DonorTools object at 0x1063d6d30>}

students/jesse_miller/session09/cli_main.py
```diff
@@ -0,0 +1,180 @@
...
+        headers = ["Donor Name", "Total Given", "Times Donated", "Average Gift"]
+        print(f"\n{'-'*80}\n{{:17}} | {{:<19}} | {{:<15}} | {{:<19}}\n{'-'*80}"\
+        .format(headers[0], headers[1], headers[2], headers[3]))
+
+        donor_data = alms.create_report()
+        for row in donor_data:
+            print('{:17} | ${:<18,.2f} | {:<15} | ${:<16,.2f}'.format\
+            (row[0], row[1], row[2], row[3]))
+        print(f"{'-'*80}\n")
+
```
ldconejo: Try not to use _create_report_ and instead generate the report directly here by pulling _alms.donors_dict_:

```
    @staticmethod
    def print_report():
        """
        Prints a summary report of all donors.
        """
        headers = ["Donor Name", "Total Given", "Times Donated", "Average Gift"]
        print(f"\n{'-'*80}\n{{:17}} | {{:<19}} | {{:<15}} | {{:<19}}\n{'-'*80}"\
        .format(headers[0], headers[1], headers[2], headers[3]))

        donor_data = alms.donors_dict
        for donor in donor_data:
             print('{:17} | ${:<18,.2f} | {:<15} | ${:<16,.2f}'.format\
            (donor_data[donor].name, sum(donor_data[donor].donations), len(donor_data[donor].donations), sum(donor_data[donor].donations)/len(donor_data[donor].donations)))           
        print(f"{'-'*80}\n")
```

UncleanlyCleric: That makes more sense.  Thanks!

students/jesse_miller/session09/donor_models.bak.py
```diff
@@ -0,0 +1,136 @@
...
+    Find a donor
+    List donors
+    Donor reports
+    Data saving and reloading.
+    '''
+    def __init__(self):
+        '''
+        Initializing and building the default dict
+        '''
+        self.donors_dict = {'Robert Smith': [435.56, 125.23, 357.10],
```
ldconejo: Remove loading sample data for now and try adding new donors manually. The problem with this initialization is that it will cause your _donors_dict_ to have two types of data:

- List of donations (for the sample donors added during initialization).
- Donor objects (added when creating new donors).

ldconejo:
Hi Jesse,

Looks pretty good. I added some comments and suggested changes to your code. With those changes, I was able to add new donors and create report showing them. 

-Luis

students/jesse_miller/session09/cli_main.py
```diff
@@ -0,0 +1,182 @@
...
+    def donor_mail_choice():
+        '''
+        This section allows the user to mail a donor
+        '''
+        donor = ''
+        MailRoom.list_donors()
+        try:
+            donor = str(input('Who would you like to mail (all for all): '))
+            if alms.find_donor(donor):
+                MailBox.mail_send(donor)
```
ldconejo: Change this one to take _alms.donor_dict[donor]_

students/jesse_miller/session09/cli_main.py
```diff
@@ -0,0 +1,182 @@
...
+            if donor == 'all':
+                MailBox.mail_send_all()
```
ldconejo: Change this one to take _alms.donor_dict_

PR #183, opened 2019-03-14
Circle exercise

mjchang14:
The steps and output examples in this assignment were very helpful in conceptualizing and designing the tests.

### DIFF
--- a/students/mjchang/session08/circle.py
+++ b/students/mjchang/session08/circle.py
@@ -58,5 +58,38 @@ class Circle(object):
             return self.radius * val
         except TypeError:
             rmul(self, val)
-            
+
+
+#### Step 8 - compare circles ####
+    def __lt__(self, other):
+        return self.radius < other.radius
+
+    def __eq__(self, other):
+        return self.radius == other.radius
+
+    def __imul__(self, val):
+        return self.radius * val
+
+    def __iadd__(self, other):
+        return self.radius + other.radius
+
+
+#### Step 9 - create a sphere ####
+class Sphere(Circle):
+    def __str__(self):
+        return "Sphere with radius: {}".format(str(self.radius))
+
+    def __repr__(self):
+        return "Sphere({})".format(self.radius)
+
+    @property
+    def volume(self):
+        return 4/3 * math.pi * self.radius**3
+
+    @property
+    def area(self):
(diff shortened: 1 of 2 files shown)


students/mjchang/session08/circle.py
```diff
@@ -0,0 +1,95 @@
...
+    def __repr__(self):
+        return "Circle({})".format(self.radius)
+
+
+#### Step 7 - add numerical protocol ####
+    def __add__(self, other):
+        return self.radius + other.radius
+
+    def __rmul__(self, val):
+        return self.radius * val
```
ldconejo: Consider adding a test for this one to increase your overall coverage.

students/mjchang/session08/circle.py
```diff
@@ -0,0 +1,95 @@
...
+
+#### Step 8 - compare circles ####
+    def __lt__(self, other):
+        return self.radius < other.radius
+
+    def __eq__(self, other):
+        return self.radius == other.radius
+
+    def __imul__(self, val):
+        return self.radius * val
```
ldconejo: Consider adding a test for this one to increase your overall coverage.

students/mjchang/session08/circle.py
```diff
@@ -0,0 +1,95 @@
...
+    def __str__(self):
+        return "Circle with radius: {}".format(str(self.radius))
+
+    def __repr__(self):
+        return "Circle({})".format(self.radius)
+
+
+#### Step 7 - add numerical protocol ####
+    def __add__(self, other):
+        return self.radius + other.radius
```
ldconejo: This works, but the intent of the assignment is for arithmetic operations between Circle objects to return Circle objects as well. In this case, for example, adding Circle(4) + Circle(6) should return a Circle(10) object instead of the integer 10.

Following a similar logic, 2*Circle(6) shoudl return a Circle(12) object.

Luis

ldconejo: Please implement the same changes for other math operations in the Circle class.

ldconejo:
Hi Marisa,

Coverage is at 88%, there are a couple of additional tests you can add to increase that past 90%. On arithmetic operations, please modify them to result in new Circle instances, instead of integers.

Regards,

Luis

PR #184, opened 2019-03-16
Updated Circle exercise - question

mjchang14:
Hi Luis and Michael,

I made the changes to the math operations as required. However, when I modified the tests to include the math operations my coverage dropped from 88% to 79%. Other than reverting back to my original submission, I can't figure out how to increase the coverage. Can you take a look and point me in the right direction if you see anything?

Thanks!


ldconejo:
Hi Marisa,

I just tested coverage for your updated code, it's at 92%. Try this command:

`$ python -m coverage run --include=circle.py -m pytest test_circle.py`

Then this one to get an html report:

`$ python -m coverage html`

This will generate a folder called _htmlcov_. The file _circle_py.html_ will show you what code is not being covered by your unit testing (only two lines of code are currently not covered).

Good job on the changes!

-Luis

PR #185, opened 2019-03-16
Assignment 9 - Mailroom OO

sdlipp:
Adding 3 files for assignment 9:
user interface code
donor db and logic
test module


students/ScottL/session09/cmi_main.py
```diff
@@ -0,0 +1,155 @@
...
+                print(name)
+        else:
+            print("\n\tThe name list is empty.")
+
+    def validate_name(self):
+        """
+        Process the expected user input
+        :return: 2-element tuple containing the first and last names
+        """
+        first = self.get_donor_name("FIRST")
```
mdgregor: you could have gotten the full name of the donor, and then run your logic for `quit` and `list` and then else return the name.
```
name = self.get_donor_name("first and last").lower()   # you could just ask for name instead of first/last in your        text

if name == "list":
    self.display_donor_list()
elif name == "quit":
    return
else:
    return name
```

students/ScottL/session09/cmi_main.py
```diff
@@ -0,0 +1,155 @@
...
+                    return first, last
+
+    @staticmethod
+    def get_donation_amount():
+        """
+        Receive and process input from user on the donation amount.
+        :return: float-type number of donation amount
+        """
+        prompt = "Enter the amount of the donation: "
+        amount = input(prompt)
```
mdgregor: no reason for these to be two separate lines.  just do `amount = input("Enter the amount of the donation: ")`

students/ScottL/session09/cmi_main.py
```diff
@@ -0,0 +1,155 @@
...
+        value_error_message = "\nThe entry for 'amount' count was not a valid number. " \
+                              "Please try again."
+        if amount.strip().lower() not in ("quit", ""):
+            try:
+                amount = "".join(amount.split(","))
+                amount = float(amount.strip("$"))
+            except ValueError:
+                print(value_error_message)
```
mdgregor: this should be wrapped in a while loop, otherwise it just returns nothing.

students/ScottL/session09/cmi_main.py
```diff
@@ -0,0 +1,155 @@
...
+        else:
+            print("\t\t\t*** No data exists to display ***")
+
+    @staticmethod
+    def email():
+        """Write text(s) to file simulating email(s) thanking donor(s)."""
+        emails = md.DonorCollection().group_email()
+        for email in emails:
+            try:
+                with open(f"{email}.txt", "w") as file:
```
mdgregor: `file` is a python keyword.  You should change it to `my_file` or something else.

students/ScottL/session09/cmi_main.py
```diff
@@ -0,0 +1,155 @@
...
+                          "please input a positive number for the donation.")
+                else:
+                    return amount
+
+    def new_donation(self):
+        """Organize process flow for calling and validating user input for new donations"""
+        try:
+            first, last = self.validate_name()
+        except TypeError:
+            self.main_menu()
```
mdgregor: you want to return, not call the main_menu as you'll get some weird exiting logic and you try to return back up the stack.

students/ScottL/session09/mailroom_donor.py
```diff
@@ -0,0 +1,142 @@
...
+        :param amount: float, donation amount
+        :return string, personalized 'thank-you' email
+        """
+        self.donations.append(amount)
+        return self.solo_email(amount)
+
+    @property
+    def count(self):
+        """:return integer, the number of donations for a particular donor."""
+        return len(self.donations)
```
mdgregor: could you have made this a `def __len__(self):` function?

mdgregor:
Here is a good example of how your classes should operate. https://repl.it/@naaleksandrova/Lesson9Example

PR #187, opened 2019-03-17
Session 09

dac2229:
Final assignment

### DIFF
--- a/students/daniel_carrasco/session09/cli_main.py
+++ b/students/daniel_carrasco/session09/cli_main.py
@@ -51,7 +51,7 @@ def thankyou():
                     3 - Enter New Name.\n\
                     4 - Quit submenu\n'))
             if choice == 1:
-                print(dm.Donor.showallnames())
+                print(ddb.Donor_Database.showallnames())
             if choice == 2:
                 key = input('Enter full name\n')
                 value = dm.Donor(key).donations
@@ -80,7 +80,7 @@ def get_report():
             'Donation ($)',
             'Amount',
             'Average ($)'))
-    print(dm.Donor.showavg())
+    print(ddb.Donor_Database.showavg())
     for key, value in ddb.Donor_Database.data.items():
         report_list.append(
             f'{key.title():>15}{value[0]:>15.2f}{value[1]:>15}{ddb.Donor_Database.data[key][0]/ddb.Donor_Database.data[key][1]:>15.2f}')
@@ -114,7 +114,7 @@ def addnew():
                     3 - Enter Donation.\n\
                     4 - Quit submenu\n'))
             if choice == 1:
-                print(dm.Donor.showallnames())
+                print(ddb.Donor_Database.showallnames())
             if choice == 2:
                 key = input('Enter full name\n')
                 dm.Donor(key).donations = 0
--- a/students/daniel_carrasco/session09/cli_main.py
+++ b/students/daniel_carrasco/session09/cli_main.py
@@ -1,0 +1,134 @@
+import donor_models as dm
+import donor_database as ddb
+import os
+import sys
+import pytest
+
+donors = {'Art Bart': [1000, 1], 'Harry Scary': [50, 5], 'Hay Boo': [50000, 3]}
+for keys in donors:
+    dm.Donor(keys).donations = donors[keys][0]
+    dm.Donor(keys).counter = donors[keys][1]
+
+path = os.getcwd()
+
+
+def main():
+
+    while True:
+        try:
+            user_choice = int(input('\nChoose an action(1 - 4):\n\n\
+                    1 - Send a Thank You to a single donor.\n\
+                    2 - Create a Report.\n\
+                    3 - Send letters to all donors.\n\
+                    4 - Add new donor.\n\
+                    5 - Add new donation.\n\
+                    6 - Quit\n'))
+            arg_dict = {
+                1: thankyou,
+                2: report,
+                3: letter,
+                4: addnew,
+                5: addnew,
+                6: quit}
+
+            if arg_dict[user_choice] == 'quit':
+                sys.exit()
+            else:
+                arg_dict[user_choice]()
+        except ValueError:
+            print("Input must be an integer, try again.")
+        except KeyError:
+            print('Choice must be a menu input (1-6)')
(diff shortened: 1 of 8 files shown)


students/daniel_carrasco/session09/cli_main.py
```diff
@@ -0,0 +1,134 @@
...
+                    4 - Add new donor.\n\
+                    5 - Add new donation.\n\
+                    6 - Quit\n'))
+            arg_dict = {
+                1: thankyou,
+                2: report,
+                3: letter,
+                4: addnew,
+                5: addnew,
+                6: quit}
```
mdgregor: shouldn't this be `6: "quit"}` ?

students/daniel_carrasco/session09/cli_main.py
```diff
@@ -0,0 +1,134 @@
...
+
+            if arg_dict[user_choice] == 'quit':
+                sys.exit()
+            else:
+                arg_dict[user_choice]()
+        except ValueError:
+            print("Input must be an integer, try again.")
+        except KeyError:
+            print('Choice must be a menu input (1-6)')
+        continue
```
mdgregor: no need for the `continue` here.

students/daniel_carrasco/session09/cli_main.py
```diff
@@ -0,0 +1,134 @@
...
+                value = dm.Donor(key).donations
+                [print(row) for row in write_letter(key, value)]
+            if choice == 3:
+                key = input('Enter full name\n')
+                dm.Donor(key)
+            if choice == 4:
+                main()
+        except ValueError:
+            print("Input must be an integer, try again.")
+            continue
```
mdgregor: not needed here either.

students/daniel_carrasco/session09/donor_database.py
```diff
@@ -0,0 +1,47 @@
...
+        return self.data[self.name][0]
+
+    def counter(self, count=None):
+        if count is not None:
+            current_donation = self.data[self.name][0]
+            self.data.update({self.name: [current_donation, count]})
+        return self.data[self.name][1]
+
+    @staticmethod
+    def showallnames():
```
mdgregor: `show_all_names`

students/daniel_carrasco/session09/donor_database.py
```diff
@@ -0,0 +1,47 @@
...
+        return [Donor_Database.data[i][0]
+                for i in iter(Donor_Database.data)]
+
+    @staticmethod
+    def showallcounts():
+        return [Donor_Database.data[i][1]
+                for i in iter(Donor_Database.data)]
+
+    @staticmethod
+    def showavg():
```
mdgregor: all of the above functions need to have `_` between their words.

students/daniel_carrasco/session09/donor_database.py
```diff
@@ -0,0 +1,47 @@
+class Donor_Database:
```
mdgregor: classes should be `DonorDatabase`

students/daniel_carrasco/session09/donor_database.py
```diff
@@ -0,0 +1,47 @@
...
+    data = {}
+
+    def __init__(self, name=""):
```
mdgregor: I would think you would always want `name` to be required for the instantiation of the class.

students/daniel_carrasco/session09/donor_database.py
```diff
@@ -0,0 +1,47 @@
...
+        return self.data[self.name][1]
+
+    @staticmethod
+    def showallnames():
+        return [i for i in iter(Donor_Database.data)]
+
+    @staticmethod
+    def showalldonations():
+        return [Donor_Database.data[i][0]
+                for i in iter(Donor_Database.data)]
```
mdgregor: This should be part of a DonorCollection class.  A single donor instance shouldn't hold multiple donors.

students/daniel_carrasco/session09/donor_database.py
```diff
@@ -0,0 +1,47 @@
...
+
+    @staticmethod
+    def showallcounts():
+        return [Donor_Database.data[i][1]
+                for i in iter(Donor_Database.data)]
```
mdgregor: same with this one.

students/daniel_carrasco/session09/donor_database.py
```diff
@@ -0,0 +1,47 @@
...
+                self.data[name]
+            except KeyError:
+                self.data.update({name: [0, 0]})
+
+    def donations(self, amount=0):
+        current_donation = self.data[self.name][0]
+        current_count = self.data[self.name][1]
+        if amount == 0:
+            self.data.update(
+                {self.name: [amount + current_donation, 1]})
```
mdgregor: `list` objects that you have here have a length you can derive by calling `len()` on them, so you don't need to keep a count of donations as part of your data structure.

mdgregor: example:
```
>>> donations = {"john": [123.00, 3123.02, 112.00]}
>>> len(donations["john"])
3
```

students/daniel_carrasco/session09/donor_database.py
mdgregor: this isn't how you want to iterate through a dictionary.  Your treating it like a `list`.  You want to do:
```
return Donor_Database.data.values()
```

students/daniel_carrasco/session09/donor_database.py
mdgregor:
```
return [len(values) for values in Donor_Database.data.values()]
```

students/daniel_carrasco/session09/cli_main.py
```diff
@@ -0,0 +1,134 @@
...
+            arg_dict = {
+                1: thankyou,
+                2: report,
+                3: letter,
+                4: addnew,
+                5: addnew,
+                6: quit}
+
+            if arg_dict[user_choice] == 'quit':
+                sys.exit()
```
mdgregor: you can actually get rid of this `if` statement if you keep your `quit` without the quotes and it will execute `quit()` which is a python built in and will do the same as `sys.exit()`

PR #189, opened 2019-03-18
Submitting Lesson 09 - Mailroom OO

jroam142:


students/jeremy_m/lesson09_exercise/mailroom_oo/cli_main.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+class DonorLetters():
+    """ 
+    Handles the creation of txt files formatted as thank you letters for each
+    donor in the donors dict.
+    """
+
+    def __init__(self):
+        pass
```
mdgregor: if your not going to use the function there is no reason to define it.

PR #192, opened 2019-03-18
Woosukjeung patch fixed

woosukjeung:
i've just figured out what was wrong with my pytest i'll submit the rest as soon as possible
9368351


woosukjeung:
saw all the submission has be processed i will close this pull to reduce redundancy

